PR #7, opened 2020-05-29
Added support for queries of the form EXEC <macro_name>;

AndrewPochapsky:


babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code EXEC} statement.
+ */
+public class SqlExecMacro extends SqlDdl implements SqlExecutableStatement {
```
AndrewPochapsky, line 36: Wasn't sure which class to extend from here. I wasn't able to access SqlNode's constructor so I went with SqlDdl.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,32 @@ SqlCreate SqlCreate() :
...
+ */
+SqlNode SqlExec() :
+{
+    final Span s;
+	SqlNode macro;
+	SqlNode source;
+}
+{
+	/** Macro has the same domain as Table so TableRef() works */
+	<EXEC> macro = TableRef() { s = span(); }
```
AndrewPochapsky, line 3974: I think this is fine but I looked at TableRef() and it seems to do way more than I need, is there an alternative?

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
dasch-1: Maybe we could try extending from SqlCall?

https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/SqlCall.java

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,32 @@ SqlCreate SqlCreate() :
...
 </#if>
-
+<#if parser.execStatementParserMethods?size != 0>
+/**
+ * Parses an EXEC statement
+ */
+SqlNode SqlExec() :
+{
+    final Span s;
+	SqlNode macro;
```
dasch-1, line 3969: nit: indentation

AndrewPochapsky: Odd, it looks fine locally.

core/src/main/codegen/templates/Parser.jj
dasch-1: Agreed. We should try something a bit simpler like CompoundIdentifier().

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,32 @@ SqlCreate SqlCreate() :
...
+	(
+<#-- additional literal parser methods are included here -->
+<#list parser.execStatementParserMethods as method>
+		source = ${method}(s, macro)
+	|
+</#list>
+        source = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
+	)
+	{
+		return source;
```
dasch-1, line 3984: Can we move this logic into the execStatementParserMethod itself rather than outside? Note that methods defined in the parserImpls.ftl file still have access to all the functions in Parser.jj.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,32 @@ SqlCreate SqlCreate() :
...
+{
+	/** Macro has the same domain as Table so TableRef() works */
+	<EXEC> macro = TableRef() { s = span(); }	
+	(
+<#-- additional literal parser methods are included here -->
+<#list parser.execStatementParserMethods as method>
+		source = ${method}(s, macro)
+	|
+</#list>
+        source = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
```
dasch-1, line 3981: Could we get away with something simpler like ParenthesizedQueryOrCommaList()?

core/src/main/codegen/templates/Parser.jj
dasch-1: Could it be related to using tabs vs. spaces in your editor?

AndrewPochapsky: Looks like it, should be resolved in next commit

core/src/main/codegen/templates/Parser.jj
AndrewPochapsky: In this case should we even have the source = ... line if we don't have any default logic?

AndrewPochapsky: Since the whole code block is wrapped in an <if>,  by definition there should be a method to call so I feel like the default source setting is unnecessary as it shouldn't ever happen anyways.

PR #8, opened 2020-05-29
Added support for SEL keyword with same functionality as SELECT

michaelvdow:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -1259,8 +1259,10 @@ SqlSelect SqlSelect() :
 {
     (
         <SELECT>
+    <#if parser.includeSelKeyword>
```
dasch-1, line 1262: Beautiful.

dasch-1:
Nice work! My only feedback would be to squash all commits into a single one per round of review.

PR #9, opened 2020-05-29
Added SEL keyword support with same functionality as SELECT

michaelvdow:


babel/src/main/codegen/config.fmpp
```diff
@@ -32,6 +32,7 @@ data: {
       "IF"
       "SEMI"
       "VOLATILE"
+      "SEL"
```
dasch-1, line 35: nit: let's keep these in alphabetical order.

dasch-1, line 35: nit: let's keep these in alphabetical order.

AndrewPochapsky: Just wanted to hop in and inquire if putting these in reserved keywords instead of non-reserved could have any issues down the line? For example, what if a user in a dialect that doesn't have SEL uses it for an attribute name or something?

michaelvdow: I changed the order of the keywords, but I am also wondering now if we should put it as a non-reserved keyword since only some dialects use it

dasch-1: Per our conversation today, nonReservedKeywords SGTM

michaelvdow: Made a commit to move SEL to nonReservedKeywords

PR #10, opened 2020-05-29
Added support for queries of the form EXEC <macro_name>;

AndrewPochapsky:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -1015,6 +1015,10 @@ SqlNode SqlStmt() :
         stmt = SqlCreate()
     |
 </#if>
+<#if parser.execStatementParserMethods?size != 0>
+		stmt = SqlExec()
+	|
```
dasch-1, line 1020: nit: indentation

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,34 @@ SqlCreate SqlCreate() :
...
+SqlNode SqlExec() :
+{
+	final Span s;
+	SqlNode macro;
+	SqlNode source;
+}
+{
+	(
+		<EXEC> 
+		|
```
dasch-1, line 3975: nit: I think the style is to align the | with the ( and ) indentation level.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3955,7 +3959,34 @@ SqlCreate SqlCreate() :
...
+	SqlNode macro;
+	SqlNode source;
+}
+{
+	(
+		<EXEC> 
+		| 
+		<EXECUTE>
+	)
+	macro = CompoundIdentifier() { s = span(); }
```
dasch-1, line 3978: Should extracting the macro be moved into the SqlExecMacro method in parserImpls.ftl?

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code EXEC} statement.
+ */
+public class SqlExecMacro extends SqlCall implements SqlExecutableStatement {
+  public static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("EXECUTE", SqlKind.OTHER);
+
+  private final SqlNode name;
```
dasch-1, line 40: Can this be of type SqlIdentifier?

core/src/main/codegen/templates/Parser.jj
AndrewPochapsky: Hard to say, not sure what other queries will start with EXEC. If there aren't any it can be moved in.

dasch-1: From purely reading the function names alone (SqlExec vs. SqlExecMacro), it would seem most fitting to move the macro processing into the latter. That leaves future flexibility for us to handle EXEC/EXECUTE statements that don't have macros in this outer function.

AndrewPochapsky: Ah that makes sense, will make the change.

PR #11, opened 2020-05-29
[CALCITE] Add parsing support for ON COMMIT and WITH DATA in CREATE TABLE statements.

dasch-1:


babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateTable.java
```diff
@@ -42,19 +42,30 @@
...
+  public final OnCommitType onCommitType;
 
   private static final SqlOperator OPERATOR =
       new SqlSpecialOperator("CREATE TABLE", SqlKind.CREATE_TABLE);
 
   /** Creates a SqlCreateTable. */
   public SqlCreateTable(SqlParserPos pos, boolean replace, SetType setType, boolean isVolatile,
       boolean ifNotExists, SqlIdentifier name, SqlNodeList columnList, SqlNode query) {
+    this(pos, replace, setType, isVolatile, ifNotExists, name, columnList, query,
+        /*withData=*/false, /*onCommitType=*/OnCommitType.UNSPECIFIED);
```
AndrewPochapsky, line 55: What does the /*var_name*/ do here? Is that how optional parameters are set with java?

dasch-1: No, it's just a helpful hint to a reader of what the unnamed parameter is. It's typically used wherever literal parameters that have no implied context (e.g. false or 0) are passed into a function.

It's a Google style thing and probably used elsewhere as well.

dasch-1: I guess I could have left off the /*onCommitType=*/ one since the enum itself contains context. But being too explicit never hurts :)

AndrewPochapsky: That makes a lot of sense, thanks!

PR #12, opened 2020-06-01
[CALCITE] Added support for queries of form UPDATE <table_1> FROM <table_2>...

AndrewPochapsky:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -1655,7 +1657,12 @@ SqlNode SqlUpdate() :
         }
     ]
     [ [ <AS> ] alias = SimpleIdentifier() ]
-    <SET> id = SimpleIdentifier() {
+    (
+        <FROM> secondTable = TableRef()
+        [ [ <AS> ] secondAlias = SimpleIdentifier() ]
+    )?
```
AndrewPochapsky, line 1663: I ultimately wasn't sure how to do this using an injected method but I feel like this solution is fairly clean.

PR #13, opened 2020-06-01
[CALCITE] Added support for queries of form UPDATE <table_1> FROM <table_2>

AndrewPochapsky:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -1661,7 +1663,12 @@ SqlNode SqlUpdate() :
         }
     ]
     [ [ <AS> ] alias = SimpleIdentifier() ]
-    <SET> id = SimpleIdentifier() {
+    (
+        <FROM> secondTable = TableRef()
+        [ [ <AS> ] secondAlias = SimpleIdentifier() ]
+    )?
```
AndrewPochapsky, line 1669: I ultimately wasn't sure how to do this using an injected method but I feel like this solution is fairly clean.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -304,6 +304,20 @@ private void checkParseInfixCast(String sqlType) {
...
   }
 
+  @Test public void testUpdateFromTable() {
+    final String sql = "update foo from bar set foo.x = bar.y, foo.z = bar.k";
+    final String expected = "UPDATE `FOO` FROM `BAR` SET `FOO`.`X` = `BAR`.`Y`, "
+        + "`FOO`.`Z` = `BAR`.`K`";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testUpdateFromTableWithAlias() {
```
dasch-1, line 314: Should we slightly modify this test to have both tables be aliased?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1661,7 +1663,12 @@ SqlNode SqlUpdate() :
...
+    /** CompoundIdentifier() can read statements like FOO.X, SimpleIdentifier() is unable to do this */
```
dasch-1, line 1670: nit: you can can just use a regular "//" comment here.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1752,7 +1759,7 @@ SqlUpdate WhenMatchedClause(SqlNode table, SqlIdentifier alias) :
     )*
     {
         return new SqlUpdate(s.addAll(updateExprList).pos(), table,
-            updateColumnList, updateExprList, null, null, alias);
+            updateColumnList, updateExprList, null, null, alias, /*secondTable=*/ null, /*secondAlias=*/ null);
     }
```
dasch-1, line 1763: nit: Shorten line to be <=80 characters.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1683,7 +1690,7 @@ SqlNode SqlUpdate() :
     {
         return new SqlUpdate(s.addAll(targetColumnList)
             .addAll(sourceExpressionList).addIf(condition).pos(), table,
-            targetColumnList, sourceExpressionList, condition, null, alias);
+            targetColumnList, sourceExpressionList, condition, null, alias, secondTable, secondAlias);
```
dasch-1, line 1693: nit: Shorten line to be <=80 characters.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1661,7 +1663,12 @@ SqlNode SqlUpdate() :
         }
     ]
     [ [ <AS> ] alias = SimpleIdentifier() ]
-    <SET> id = SimpleIdentifier() {
+    (
```
dasch-1, line 1666: We only want this new logic in Babel mode. So suggest moving this to Babel's parserImpls.ftl.

AndrewPochapsky: How would I go about doing this? Is there a way to inject a code snippet rather than a whole method?

dasch-1: Yep, here's an example from Michael's recent change to add support for SEL:
https://github.com/googleinterns/calcite/blob/master/core/src/main/codegen/templates/Parser.jj#L1266

AndrewPochapsky: Ah I see, so I should just add a boolean flag and wrap the snippet in an if?

dasch-1: Yeah I think that could be one way to go about it. Something like parser.allowUpdateFromTable?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1661,7 +1663,16 @@ SqlNode SqlUpdate() :
...
     [ [ <AS> ] alias = SimpleIdentifier() ]
-    <SET> id = SimpleIdentifier() {
+<#if parser.allowUpdateFromTable>
+    (
+        <FROM> secondTable = TableRef()
+        [ [ <AS> ] secondAlias = SimpleIdentifier() ]
+    )?
+</#if>
+    /* CompoundIdentifier() can read statements like FOO.X, SimpleIdentifier()
+        is unable to do this
```
dasch-1, line 1673: nit: Indentation looks a little weird.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3975,8 +3988,8 @@ SqlNode SqlExec() :
 }
 {
     (
-        <EXEC> 
-    | 
+        <EXEC>
```
dasch-1, line 3991: Can we pull in these changes from master into this branch?

PR #14, opened 2020-06-01
[CALCITE] Added test case for SEL functionality

michaelvdow:


dasch-1:
I think you also need to remove SEL from here:
https://github.com/googleinterns/calcite/blob/master/babel/src/main/codegen/config.fmpp#L356

PR #15, opened 2020-06-01
fixed trailing whitespace in a few files

AndrewPochapsky:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -160,17 +160,17 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     }
 }
 
-/** 
-    Reason for having this is to be able to return the SqlExecMacro class since Parser.jj does not have it as a reference. 
+/**
+    Reason for having this is to be able to return the SqlExecMacro class since Parser.jj does not have it as a reference.
```
dasch-1, line 164: While we're at it, let's also wrap this comment so it's less than ~90ish lines.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -160,17 +160,20 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     }
 }
 
-/** 
-    Reason for having this is to be able to return the SqlExecMacro class since Parser.jj does not have it as a reference. 
-    This can also likely be extended to accomodate optional parameters later if need be.
+/**
+    Reason for having this is to be able to return the SqlExecMacro
+    class since Parser.jj does not have it as a reference.
```
dasch-1, line 165: Heh, that seems a little too short :)

Standard style is a max of 80 characters per line. But Calcite developers have seemed to go up to 90 characters. So anything in that range should be fine.

I would say going with 80 characters could be a safe bet and a good style habit to use even outside this project. In fact, in vim you can turn on a visual line length marker at 80 characters so you can easily see lines that go over this boundary. You might even be able to enable a mode which automatically wraps lines for you at the 80 character marker.

PR #16, opened 2020-06-01
[CALCITE] Added inline MOD operator support of form "x MOD y"

michaelvdow:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -342,4 +342,9 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testInlineModSyntax() {
```
dasch-1, line 345: nit: Suggest we add a test for both floating point and integer-only arguments to MOD.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -286,3 +286,27 @@ void InfixCast(List<Object> list, ExprContext exprContext, Span s) :
         list.add(dt);
     }
 }
+
+// Parses inline MOD expression of form "x MOD y"
```
dasch-1, line 290: nit: Suggest adding a comment indicating what types x and y can be.

PR #18, opened 2020-06-02
[CALCITE] Added support for FORMAT in CAST expression

michaelvdow:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -342,4 +342,9 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testCastFormat() {
```
dasch-1, line 345: This is a great test case. But I suggest we add at least one other type as well. For example, something like:

CAST('2020-06-02' AS DATE FORMAT 'yyyy-mm-dd')

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -118,12 +118,18 @@ public RelDataType inferReturnType(
...
   public String getSignatureTemplate(final int operandsCount) {
-    assert operandsCount == 2;
-    return "{0}({1} AS {2})";
+    switch (operandsCount) {
+    case 2:
+      return "{0}({1} AS {2})";
+    case 3:
+      return "{0}({1} AS {2} FORMAT {3})";
+    }
+    assert false;
```
dasch-1, line 127: nit: Suggest putting both of these lines in the default case.

PR #19, opened 2020-06-02
[CALCITE] Added support for UPPERCASE and CASESPECIFIC column level attributes for CREATE TABLE expression

AndrewPochapsky:


core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -153,13 +167,42 @@ public Boolean getNullable() {
...
+  public SqlDataTypeSpec withUppercase(Boolean uppercase) {
+    if (Objects.equals(uppercase, this.uppercase)) {
+      return this;
+    }
+    return new SqlDataTypeSpec(typeNameSpec, timeZone, nullable, uppercase,
+        caseSpecific, getParserPosition());
+  }
+
+  /** Returns a copy of this data type specification with a given
+   * nullability. */
```
michaelvdow, line 199: I think this comment could be updated for case specific instead of nullability

AndrewPochapsky: You're correct, darn copy and paste!

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -133,23 +133,83 @@ SqlNodeList ExtendColumnList() :
     }
 }
 
+boolean ProcessNull() :
```
dasch-1, line 136: /ProcessNull/IsNullable/

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -133,23 +133,83 @@ SqlNodeList ExtendColumnList() :
...
+            return false;
+        }
+    |
+        <NULL> {
+            return true;
+        }
+    )
+}
+
+boolean ProcessCaseSpecific() :
```
dasch-1, line 152: /ProcessCaseSpecific/IsCaseSpecific/

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -133,23 +133,83 @@ SqlNodeList ExtendColumnList() :
...
+            return false;
+        }
+    |
+        <CASESPECIFIC> {
+            return true;
+        }
+    )
+}
+
+boolean ProcessUpperCase() :
```
dasch-1, line 168: /ProcessUpperCase/IsUpperCase/

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -133,23 +133,83 @@ SqlNodeList ExtendColumnList() :
...
+    Boolean caseSpecific = null;
     final Span s = Span.of();
 }
 {
     id = CompoundIdentifier()
     type = DataType()
-    [
-        <NOT> <NULL> {
-            nullable = false;
+    //This acts as a loop to check which optional parameters have been specified
```
dasch-1, line 196: nit: Missing a space between slash and first word.

dasch-1, line 196: nit: Suggest ending comments with punctuation.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -308,6 +308,23 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testCreateTablePermutedColumnLevelAttributes() {
```
dasch-1, line 311: We should add test cases where only a subset of the attributes are explicitly specified (e.g. just "uppercase" or "not uppercase", but none of the others).

core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -83,7 +89,8 @@
   public SqlDataTypeSpec(
       final SqlTypeNameSpec typeNameSpec,
       SqlParserPos pos) {
-    this(typeNameSpec, null, null, pos);
+    this(typeNameSpec, /*timeZone=*/ null, /*nullable=*/ null, /*uppercase=*/ null,
```
dasch-1, line 92: Thanks for adding these!

core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -97,27 +104,34 @@ public SqlDataTypeSpec(
...
       SqlParserPos pos) {
-    this(typeNameSpec, timeZone, null, pos);
+    this(typeNameSpec, timeZone, /*nullable=*/ null, /*uppercase=*/ null,
+        /*caseSpecific=*/ null, pos);
   }
 
   /**
    * Creates a type specification representing a type, with time zone,
-   * nullability and base type name specified.
+   * nullability, uppercase status, caseSpecific status, and base type name specified.
```
dasch-1, line 113: nit: wrap this line to 80 characters

core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -97,27 +104,34 @@ public SqlDataTypeSpec(
...
    *
    * @param typeNameSpec The type name can be basic sql type, row type,
    *                     collections type and user defined type
    * @param timeZone     Specified time zone
    * @param nullable     The nullability
+   * @param uppercase    The uppercase status
```
dasch-1, line 119: The existing param comments are admittedly already pretty weak, but I feel like we can add a little more information here :)

core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
dasch-1: Also, the indentation looks a little off.

core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -153,13 +167,42 @@ public Boolean getNullable() {
...
     if (Objects.equals(nullable, this.nullable)) {
       return this;
     }
-    return new SqlDataTypeSpec(typeNameSpec, timeZone, nullable, getParserPosition());
+    return new SqlDataTypeSpec(typeNameSpec, timeZone, nullable, uppercase,
+        caseSpecific, getParserPosition());
+  }
+
+  /** Returns a copy of this data type specification with a given
+   * uppercase status. */
```
dasch-1, line 189: Indentation?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dasch-1: With these new tests, we can verify that attributes which aren't specified are not written out during unparse.

PR #21, opened 2020-06-02
Revert "[CALCITE] Added support for UPPERCASE and CASESPECIFIC column level attributes for CREATE TABLE expression"

AndrewPochapsky:
Reverts googleinterns/calcite#19


AndrewPochapsky:
Accidentally merged this in

PR #22, opened 2020-06-02
[CALCITE] Added support for UPPERCASE and CASESPECIFIC column level attributes for CREATE TABLE expression

AndrewPochapsky:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -133,23 +133,84 @@ SqlNodeList ExtendColumnList() :
...
+        uppercase = IsUpperCase() {
+            type = type.withUppercase(uppercase);
+        }
+    |
+        caseSpecific = IsCaseSpecific() {
+            type = type.withCaseSpecific(caseSpecific);
         }
-    ]
+    )*
+
```
dasch-1, line 211: nit: remove extra newline

PR #23, opened 2020-06-02
[CALCITE] Added support for SET TIME ZONE <time_zone_string> queries

AndrewPochapsky:


core/src/main/java/org/apache/calcite/sql/SqlSetOption.java
```diff
@@ -79,6 +79,10 @@
    * identifiers by the parser. */
   SqlNode value;
 
+  /** Controls whether the unparsed string contains an equals sign
+   * between the name and value */
```
dasch-1, line 83: nit: indentation

PR #29, opened 2020-06-03
[CALCITE] Added support for statements of form USING (var1 type1, var2 type2, ...)

AndrewPochapsky:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -4008,6 +4012,29 @@ SqlNode SqlExec() :
     }
 }
 </#if>
+<#if parser.usingStatementParserMethods?size != 0>
+/**
+ * Parses a Using statement
```
dasch-1, line 4017: nit: punctuation

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -356,6 +356,18 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testUsingRequestModifierSinglar() {
```
dasch-1, line 359: nit: Singular

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -356,6 +356,18 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "using (foo a)";
```
dasch-1, line 360: Suggest using a data type here (e.g. INTEGER or STRING) rather than "a". Same for other tests below.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -209,6 +209,17 @@ SqlNode SqlExecMacro() :
     }
 }
 
+SqlNode SqlUsingRequestModifier(Span s) :
```
ArthurInTheShell, line 212: Since this function is calling ExtendColumnList(), base on some good practice in ordering function, I suggest this function could be move upward in the code so the function call dependency would point in a downward direction.

babel/src/main/java/org/apache/calcite/sql/babel/SqlUsingRequestModifier.java
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.babel;
+
+import org.apache.calcite.jdbc.CalcitePrepare;
+import org.apache.calcite.sql.SqlCall;
```
ArthurInTheShell, line 20: While this import list looks totally fine, I would suggest you to also consider using wildcard character to avoid the long import list. Using wildcard import could reduce true dependencies and therefore would keep the module less coupled.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4033,6 +4037,29 @@ SqlNode SqlExec() :
...
+    final Span s;
+    SqlNode source;
+}
+{
+    <USING> { s = span(); }
+    (
+<#-- additional literal parser methods are included here -->
+<#list parser.usingStatementParserMethods as method>
+        source =  ${method}(s)
+        <#sep>|</#sep>
```
ArthurInTheShell, line 4055: Also a question about JavaCC and this project here. Are you expecting a use case of multiple USING statements? Since there is only one method in parser.usingStatementParserMethods , is this separation necessary? Are you expecting the number of methods to grow?

babel/src/main/java/org/apache/calcite/sql/babel/SqlUsingRequestModifier.java
AndrewPochapsky: It looks like wildcard imports do not adhere to the Google java style guide: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

core/src/main/codegen/templates/Parser.jj
AndrewPochapsky: I don't have enough domain knowledge to know if the list will grow but it is possible I suppose. I'm in the same boat as you in this case and have just been following what other methods do as a base for my designs.

babel/src/main/codegen/includes/parserImpls.ftl
AndrewPochapsky: Good suggestion, I'll keep that in mind for future features :)

babel/src/main/java/org/apache/calcite/sql/babel/SqlUsingRequestModifier.java
ArthurInTheShell: Good point! Thank you for pointing this out.

core/src/main/codegen/templates/Parser.jj
dasch-1: Since the methods are added using a list syntax in config.fmpp, it seems like best practice to write the corresponding logic in Parser.jj to handle the case of elements being added to that list in the future.

Otherwise, a developer could resonably add a new entry into the list in config.fmpp and be confused why the code no longer compiles.

PR #35, opened 2020-06-05
Added support for CREATE TABLE attribute WITH JOURNAL TABLE

michaelvdow:


babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeJournalTable.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class SqlCreateAttributeJournalTable extends SqlCreateAttribute {
+
+  private final SqlIdentifier tableName;
+
+  /**
+   * Creates a {@code SqlCreateAttributeJournalTable}.
+   *
+   * @param tableName  Name of the permanent journal table to be used
+   * @param pos  Parser position, must not be null
```
AndrewPochapsky, line 35: nit: punctuation

AndrewPochapsky: Actually never mind, I don't think we actually add periods to @param statements

michaelvdow: I think @param lines typically don't have a period at the end

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -327,6 +327,26 @@ private void checkParseInfixCast(String sqlType) {
...
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeJournalTableWithSource() {
+    final String sql = "create table foo, with journal table = baz.tbl (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, WITH JOURNAL TABLE = `BAZ`.`TBL` (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeMultipleAttrs() {
```
AndrewPochapsky, line 342: Should we also test that the order of attributes doesn't matter?

michaelvdow: That's a good point, I can add another test for that too

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeJournalTable.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  public SqlCreateAttributeJournalTable(SqlIdentifier tableName, SqlParserPos pos) {
+    super(pos);
+    this.tableName = tableName;
+  }
+
+  public void unparse(final SqlWriter writer, final int leftPrec, final int rightPrec) {
```
dasch-1, line 42: "@Override"

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -327,6 +327,34 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testTableAttributeJournalTable() {
+    final String sql = "create table foo, with journal table = baz (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, WITH JOURNAL TABLE = `BAZ` (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeJournalTableWithSource() {
```
dasch-1, line 336: Maybe the name of this test could be?

testTableAttributeJournalTableWithCompoundIdentifier()

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -327,6 +327,34 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testTableAttributeJournalTable() {
```
dasch-1, line 330: Maybe the name of this test could be?

testTableAttributeJournalTableWithSimpleIdentifier()

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dasch-1: Good catch, thanks!

PR #36, opened 2020-06-05
[CALCITE] Modify unparsing order of UPDATE statement for BIG QUERY dialect

AndrewPochapsky:
Big Query wants statements like `update abc from def set abc.x = def.y, abc.z = def.k;` to be unparsed as `update abc set abc.x = def.y, abc.z = def.k from def;`

### DIFF
--- a/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
+++ b/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
@@ -427,8 +427,8 @@ class BabelParserTest extends SqlParserTest {
   }
 
   @Test public void testUpdateFromTableWithAliasBigQuery() {
-    final String sql = "update foo as f set f.x = b.y, f.z = b.k";
-    final String expected = "UPDATE `foo` AS `f` FROM `bar` AS `b` SET `f`.`x` "
+    final String sql = "update foo as f from bar as b set f.x = b.y, f.z = b.k";
+    final String expected = "UPDATE `foo` AS `f` SET `f`.`x` "
         + "= `b`.`y`, `f`.`z` = `b`.`k` FROM `bar` AS `b`";
     bigQuerySql(sql).ok(expected);
   }
--- a/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
+++ b/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
@@ -432,6 +432,7 @@ class BabelParserTest extends SqlParserTest {
         + "= `b`.`y`, `f`.`z` = `b`.`k` FROM `bar` AS `b`";
     bigQuerySql(sql).ok(expected);
   }
+
   @Test public void testExecMacro() {
     final String sql = "exec foo";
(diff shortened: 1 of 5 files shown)


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -419,6 +419,20 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testUpdateFromTableBigQuery() {
+    final String sql = "update foo from bar set foo.x = bar.y, foo.z = bar.k";
+    final String expected = "UPDATE `foo` SET `foo`.`x` = `bar`.`y`, "
+        + "`foo`.`z` = `bar`.`k` FROM `bar`";
+    bigQuerySql(sql).ok(expected);
```
dasch-1, line 426: Rather than doing it like this, is it possible to reuse this existing pattern?

https://github.com/googleinterns/calcite/blob/master/core/src/test/java/org/apache/calcite/sql/parser/SqlParserTest.java#L8629

core/src/main/java/org/apache/calcite/sql/SqlUpdate.java
```diff
@@ -198,13 +201,8 @@ public SqlIdentifier getSourceAlias() {
       writer.keyword("AS");
       alias.unparse(writer, opLeft, opRight);
     }
-    if (sourceTable != null) {
-      writer.keyword("FROM");
```
dasch-1, line 202: Another option to using "if (dialect == BIGQUERY)" statements throughout this function would be to move all of the SqlUpdate unparsing logic into the dialect-specific files. This would mean relocating the existing implementation into the SqlDialect class. And then overriding it with the BigQuery-specific logic in the BigQuerySqlDialect class.

Here's an example of where this is being done:
https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/SqlIntervalLiteral.java#L75

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
AndrewPochapsky: Good idea, I didn't know that was possible.

core/src/main/java/org/apache/calcite/sql/SqlUpdate.java
AndrewPochapsky: I like the sound of this approach, I'll give it a try.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
AndrewPochapsky: When I do this the none of the unparsed identifiers have back ticks surrounding them. Is this expected?

dasch-1: Yeah it looks like in the test linked in the above comment, the unparsed version of that SQL string doesn't have any backticks either.

PR #37, opened 2020-06-05
[CALCITE] Added support for CREATE TABLE attribute JOURNAL

michaelvdow:


babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeJournal.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  public enum JournalType {
+    /**
+     * Maintains both types of journal images.
+     */
+    UNSPECIFIED,
+
+    /**
+     * Maintains a before change image
```
AndrewPochapsky, line 84: nit: punctuation

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeJournal.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    BEFORE,
+
+    /**
+     * Maintains an after change image
```
AndrewPochapsky, line 89: nit: punctuation

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,41 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
     { return new SqlCreateAttributeJournalTable(id, getPos()); }
 }
 
+SqlCreateAttribute CreateTableAttributeJournal() :
+{
+  JournalType journalType = JournalType.UNSPECIFIED;
+  JournalModifier journalModifier = JournalModifier.UNSPECIFIED;
+}
+{
+    [
```
dasch-1, line 245: Suggest wrapping these different clauses/cases in an outer parentheses and separated by |'s. There would be two benefits of this from a code readability/reasoning perspective:

1) IMO it's easier to reason about if there's only a single return statement in the end of the parsing function.
2) It would be easy to see at a quick glance which cases are mutually exclusive of one another.

michaelvdow: Yeah I like that better. It would make it a lot easier to read.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,39 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+{
+    (
+        (
+            <LOCAL> { journalModifier = JournalModifier.LOCAL; }
+        |
+            <NOT> <LOCAL> { journalModifier = JournalModifier.NOT_LOCAL; }
+        )
+        <AFTER> <JOURNAL> { journalType = JournalType.AFTER; }
+    |
+        [
```
dasch-1, line 253: This is looking a lot clearer already - thanks!

To decrease the amount of nesting even further, suggest removing the outer brackets and instead having an "else" case for the <NO>/<DUAL> and <BEFORE>/<AFTER> clauses below. Where in each "else" case, JournalType.UNSPECIFIED or JounralModifier.UNSPECIFIED are explicitly set.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,39 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
     { return new SqlCreateAttributeJournalTable(id, getPos()); }
 }
 
+SqlCreateAttribute CreateTableAttributeJournal() :
+{
+  JournalType journalType = JournalType.UNSPECIFIED;
```
dasch-1, line 241: Since this function contains fairly complex parsing logic, suggest not initializing these up here and instead explicitly setting the values to UNSPECIFIED below.

That would make it easier for the reader to easily identify in which cases UNSPECIFIED is being set.

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Ah I see, that would look a lot cleaner. I'll do that and remove the UNSPECIFIED from the above

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -328,6 +328,54 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testTableAttributeJournal() {
```
ArthurInTheShell, line 331: I would suggest adding "Default" in the test name so it is more clear that this test the default behavior of Journal. (same for the next two tests).

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeJournal.java
```diff
@@ -0,0 +1,120 @@
...
+      writer.keyword("DUAL");
+      break;
+    case LOCAL:
+      writer.keyword("LOCAL");
+      break;
+    case NOT_LOCAL:
+      writer.keyword("NOT");
+      writer.keyword("LOCAL");
+      break;
+    default:
```
ArthurInTheShell, line 61: You could choose to have a case for "UNSPECIFIED" case and omit the default case, since there would not be any other case. Reference from [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch). Also, you might not need the "break;" in default case.

ArthurInTheShell: This comment also applies to the default case in line71.

michaelvdow: It looks like the style guide says either is fine, so I may leave it because that's how it is in the SqlCreateTable class as well for Volatile. But I think the break could probably be removed since it doesn't do anything in this case

ArthurInTheShell: Sounds good

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
michaelvdow: Sure I could add default to that test

ArthurInTheShell: Well actually, since there are two things to be specified when you test the attribute "Journal", it might be clearer to point out what combination of JournalAttribute and JournalType you are trying to test in the test name. If you agree with me, then most of the test names are subject to change.

michaelvdow: How do you mean? For the journal modifiers, I have the modifier keyword on the end and the "local", "not local" only work with the "after" keyword so I didn't specify the type there. Is there another naming scheme you have in mind to make it clearer?

ArthurInTheShell: I would make comments in the newest version to clarify my suggestion

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -328,6 +328,54 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testTableAttributeJournalDefault() {
```
ArthurInTheShell, line 331: testTableAttributeJournalDefault can be change to testTableAttributeJournal_defaultModifier_defaultType

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -328,6 +328,54 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "create table foo, journal (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, JOURNAL (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeJournalBefore() {
```
ArthurInTheShell, line 337: testTableAttributeJournalBefore can be change to testTableAttributeJournal_defaultModifier_before

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -328,6 +328,54 @@ private void checkParseInfixCast(String sqlType) {
...
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeJournalAfter() {
+    final String sql = "create table foo, after journal (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, AFTER JOURNAL (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeJournalNoModifier() {
```
ArthurInTheShell, line 349: testTableAttributeJournalNoModifier can be change to testTableAttributeJournal_noModifier_before

AndrewPochapsky: My slight issue with using this naming convention is that it's different from the style of the other test names. Not that I dislike this proposed convention, I think it's actually quite nice, but this style difference is something I think we should consider.

michaelvdow: I do like that it specifies it more and I see that style is in this [style guide](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names), but I'm not sure if that would go against the style used in this project. Maybe @dasch-1 could clarify that

dasch-1: IMO, where there's an existing style in Calcite, we should try to match it. If there's no existing pattern to follow, however, I think going with whatever is recommended in the style guide makes sense.

ArthurInTheShell: I am not recommending a particular naming style, instead, I use underscore with the intention to show that idea of combining states together in the test name. As long as you get the gist, the actual naming style could varies

dasch-1: I think the underscores in the name is part of the style. And we should try to match the existing style. The alternative of trying to rewrite all of the existing test names in the new style seems not worth it to me :)

ArthurInTheShell: Yeah, the idea is not rewrite the existing test with the underscore style. It is from a previous conversation that I suggested to include both of JournalModifier and JournalType in the test name. Some test names in the current version is missing one of those two and that is what I think can be change. @dasch-1

michaelvdow: I can change the names so that it has modifier then type but in camel case so it matches the other tests. Something like testTableAttributeJournalNoModifierDefaultType

dasch-1: SGTM

PR #38, opened 2020-06-05
[CALCITE] Remove if from keyword to make it not quoted during unparsing

yzhvictor:
[CALCITE] Remove if from keyword to make it not quoted during unparsing


dasch-1:
Per conversation with @yzhvictor offline, we've added this logic in https://github.com/googleinterns/calcite/pull/46.

PR #42, opened 2020-06-08
[CALCITE] Added support for CREATE TABLE attribute MERGEBLOCKRATIO

michaelvdow:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,31 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+SqlCreateAttribute CreateTableAttributeMergeBlockRatio() :
+{
+    MergeBlockRatioModifier modifier = MergeBlockRatioModifier.UNSPECIFIED;
+    int ratio = 1;
+    boolean percent = false;
+}
+{
+    (
+        (
+            <DEFAULT_> { modifier = MergeBlockRatioModifier.DEFAULT; }
```
AndrewPochapsky, line 248: Is the underscore in the token supposed to be there?

michaelvdow: Yeah <DEFAULT_> maps to the "DEFAULT" token. I think just <DEFAULT> maps to something  else.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,31 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+        |
+            <NO> { modifier = MergeBlockRatioModifier.NO; }
+        )
+        <MERGEBLOCKRATIO>
+    |
+        <MERGEBLOCKRATIO> <EQ> ratio = UnsignedIntLiteral()
+        [ <PERCENT> { percent = true; } ]
+    )
+    {
+        if (ratio >= 1 && ratio <= 100) {
```
dasch-1, line 258: Suggest making this an explicit parser failure by throwing an exception, such as is done here:
https://github.com/googleinterns/calcite/blob/master/core/src/main/codegen/templates/Parser.jj#L404

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -356,6 +356,36 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "create table foo, mergeblockratio = ^-^20 (bar integer)";
+    final String expected = "(?s).*Encountered \"-\" at .*";
+    sql(sql).fails(expected);
+  }
+
+  @Test public void testTableAttributeMergeBlockRatioPercent() {
+    final String sql = "create table foo, mergeblockratio = 45 percent (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, MERGEBLOCKRATIO = 45 PERCENT (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
```
dasch-1, line 387: Should we have a test case where the ratio is outside of the bounds of 1 to 100 (but not negative)?

michaelvdow: Sure I can add that in

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Sure, I added an exception thrown for number literal out of range

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -356,6 +356,36 @@ private void checkParseInfixCast(String sqlType) {
...
+
```
dstekol, line 394: Maybe also worth adding a test to show that if you use a modifier keyword before MERGEBLOCKRATIO and an equals assignment after it, then parsing fails (like "NO MERGEBLOCKRATIO = 10" ) ?

michaelvdow: That's a good point, I can add another test for that

PR #43, opened 2020-06-08
[CALCITE] Added support for CREATE TABLE attribute CHECKSUM

michaelvdow:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -340,6 +340,24 @@ private void checkParseInfixCast(String sqlType) {
...
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeChecksumOff() {
+    final String sql = "create table foo, checksum = off (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, CHECKSUM = OFF (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
   @Test public void testTableAttributeMultipleAttrs() {
```
ArthurInTheShell, line 361: Would be nice if the multiple attributes test include checksum.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -249,6 +265,8 @@ List<SqlCreateAttribute> CreateTableAttributes() :
             e = CreateTableAttributeFallback()
         |
             e = CreateTableAttributeJournalTable()
+        |
+            e = CreateTableAttributeChecksum()
         ) { list.add(e); }
     )+
```
ArthurInTheShell, line 271: Would it be a valid query to have multiple checksums(don't know if it is)? If it is, it would be nice to also have test for multiple checksums.

michaelvdow: I believe there can only be one checksum attribute for a create table statement

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
michaelvdow: Yeah I'll add in another attribute for checksum to that test since there's only 2. I'm not sure if we want to do that with every attribute though since there's about 10 of them to add

ArthurInTheShell: I agree with you that we don't need to test every possible combination

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,22 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
     { return new SqlCreateAttributeJournalTable(id, getPos()); }
 }
 
+SqlCreateAttribute CreateTableAttributeChecksum() :
+{
+    ChecksumLevel checksumLevel;
```
dstekol, line 241: checksumLevel is a bit of a weird variable name, given that it implies that there are different levels to which this option can be set. This might make sense if the possible values were LOW/MEDIUM/HIGH like they apparently used to be, but now that this has been deprecated in favor of ON/OFF/DEFAULT it might make sense to rename this to something more boolean-sounding like checksumEnabled. That said, if you are shooting for backward compatibility, then it could be fine to keep it as is.

michaelvdow: Yeah I think checksumEnabled would sound better since the deprecated values aren't included here.

PR #44, opened 2020-06-08
[CALCITE] Fixed MERGE INTO with INSERT statement unparsing in Big Query

AndrewPochapsky:
**Original Big Query unparsing:** 
```
MERGE INTO T1 AS A
USING T2 AS B
ON A.X = B.X
WHEN MATCHED THEN UPDATE SET
    Y = B.Y
WHEN NOT MATCHED THEN INSERT (X, Y) (VALUES  (B.X, B.Y))
```

**New Big Query unparsing:**
```
MERGE INTO T1 AS A
USING T2 AS B
ON A.X = B.X
WHEN MATCHED THEN UPDATE SET
    Y = B.Y
WHEN NOT MATCHED THEN INSERT (X, Y) VALUES  (B.X, B.Y)
```

### DIFF
--- a/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
+++ b/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
@@ -544,10 +544,10 @@ class BabelParserTest extends SqlParserTest {
     final String sql = "merge into t1 a using t2 b on a.x = b.x when matched then "
         + "update set y = b.y when not matched then insert (x,y) values (b.x, b.y)";
     final String expected = "MERGE INTO t1 AS a\n"
-       + "USING t2 AS b\n"
-       + "ON (a.x = b.x)\n"
-       + "WHEN MATCHED THEN UPDATE SET y = b.y\n"
-       + "WHEN NOT MATCHED THEN INSERT (x, y) VALUES (b.x, b.y)";
+        + "USING t2 AS b\n"
+        + "ON (a.x = b.x)\n"
+        + "WHEN MATCHED THEN UPDATE SET y = b.y\n"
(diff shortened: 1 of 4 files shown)


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -539,4 +539,17 @@ private void checkParseInfixCast(String sqlType) {
     final String expected = "SELECT MOD(`FOO`, `BAR`)";
     sql(sql).ok(expected);
   }
+
+  @Test public void testMergeIntoBigQuery() {
```
dasch-1, line 543: We should add a test verifying that the old unparsing logic still works on the same example query when not in BigQuery SQL dialect.

AndrewPochapsky: I'll add the test. About the main comment: the only way that I can think of for now is putting if statements into the original file that checks the dialect. For this specific change that might actually be alright as it is quite a simple difference between the two. I can also spend a bit more time trying to explore different solutions.

dasch-1: One idea: can we leave the merge unparsing logic where it is today, but move the insert logic only into the dialect files?

AndrewPochapsky: Do you mean something like writer.getDialect().unparseInsertStatement(..) in which the default logic will be as before and BigQuery specific will be my change?

dasch-1: Yes, exactly.

PR #47, opened 2020-06-09
[CALCITE] Include parens when unparsing CURRENT_TIMESTAMP function for BigQuery dialect

AndrewPochapsky:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -554,6 +554,47 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String expected = "SELECT CURRENT_TIME";
+    sql(sql).ok(expected);
+  }
+  @Test public void testBothTimeFunctions() {
+    final String sql = "select current_time, current_timestamp";
+    final String expected = "SELECT CURRENT_TIME, CURRENT_TIMESTAMP";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testCurrentTimeStampFunctionBigQuery() {
```
michaelvdow, line 574: Would it be valid to pass in CURRENT_TIME() with parentheses and see if it unparses to the same thing? I'm not sure if the parsing logic is there to parse the parentheses or it's just supposed to unparse into BigQuery.

AndrewPochapsky: Looks like it works, I'll add additional tests to cover this.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -148,6 +155,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
     unparseFetchUsingLimit(writer, offset, fetch);
   }
 
+  @Override public void unparseSqlIdentifier(SqlWriter writer,
+      SqlIdentifier identifier, int leftPrec, int rightPrec) {
+    // Unparse the identifier as normal.
+    super.unparseSqlIdentifier(writer, identifier, leftPrec, rightPrec);
+    // If identifier.names.size() > 1 then this case is not valid as that means
```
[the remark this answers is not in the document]
AndrewPochapsky: Makes sense, will make the change.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -148,6 +155,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
...
 
+  /* Unparses the given identifier with parentheses at the end if it is
+   * required to have them. */
+  @Override public void unparseSqlIdentifier(SqlWriter writer,
+      SqlIdentifier identifier, int leftPrec, int rightPrec) {
+    super.unparseSqlIdentifier(writer, identifier, leftPrec, rightPrec);
+
+    boolean isValidIdentifier = identifier.names.size() == 1;
+    if (isValidIdentifier && IDENTIFIER_FUNCTIONS
+        .contains(identifier.names.get(0).toLowerCase(Locale.ROOT))) {
```
dstekol, line 166: Since you are manipulating lowercase/uppercase here, maybe it would make sense to add a few tests where the original input string is upper case? Right now it looks like the test input is always lower case. Don't necessarily need to make an uppercase version of each test, but maybe just one or two as a proof of concept.

AndrewPochapsky: Good point, I'll add a few tests where the function is in uppercase.

dasch-1: Case sensitivity should not matter in parsing so I suggest not including these test cases just to minimize redundant tests.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -149,6 +156,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
     unparseFetchUsingLimit(writer, offset, fetch);
   }
 
+  /* Unparses the given identifier with parentheses at the end if it is
+   * required to have them. */
+  @Override public void unparseSqlIdentifier(SqlWriter writer,
+      SqlIdentifier identifier, int leftPrec, int rightPrec) {
+    super.unparseSqlIdentifier(writer, identifier, leftPrec, rightPrec);
+
+    boolean isValidIdentifier = identifier.names.size() == 1;
```
dasch-1, line 165: If there is an invalid identifier, will this just not unparse anything? Should we be throwing an exception or unparsing just the final part of the compound identifier instead?

AndrewPochapsky: It will unparse as normal, there just won't be any parentheses.

AndrewPochapsky: And sure, I'll remove the uppercase tests.

dasch-1: Ah yes, I see that now. Sounds good.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
dstekol: Since the logic here is built on matching the identifier against the strings in IDENTIFIER_FUNCTIONS (which could fail if cases don't match), I was just envisioning a situation where due to some weird Locale behavior, the toLowerCase call doesn't change the string, and then the .contains() returns true if the original input is lower case but fails if it was upper case.

dasch-1: Hm, I see what you mean. That's a valid concern since this is a SqlIdentifier.

Ok well maybe a middle ground then where we just have a single uppercase test (e.g. just one for CURRENT_TIMESTAMP)? Along the same lines, we could have just a single test with the input query having parens (e.g. just one for CURRENT_TIMESTAMP).

The situation I'm trying to avoid is adding _too_ many tests that have a lot of coverage overlap. IMO the incremental test coverage gains can come at the cost of a bloated test file and extra burden on developers to write more tests than are really needed.

AndrewPochapsky: Okay, new changes reflect this middle ground approach.

dstekol: Agreed, no need for adding tons of uppercase tests. Just seems like the de-facto standard for these test cases is to always pass in lower case, which in certain situations (like this one) could potentially mask bugs.

PR #48, opened 2020-06-09
[CALCITE] Added support for CREATE TABLE attribute DATABLOCKSIZE

michaelvdow:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,35 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+        )
+        <DATABLOCKSIZE> { unitSize = DataBlockUnitSize.UNSPECIFIED; }
+    |
+        <DATABLOCKSIZE> <EQ> dataBlockSize = UnsignedNumericLiteral()
+        (
+            <BYTES> { unitSize = DataBlockUnitSize.BYTES; }
+        |
+            ( <KILOBYTES> | <KBYTES> ) { unitSize = DataBlockUnitSize.KILOBYTES; }
+        |
+            { unitSize = DataBlockUnitSize.UNSPECIFIED; }
```
AndrewPochapsky, line 262: Should this be DataBlocUnitSize.BYTES? I saw in the DataBlockUnitSize enum definition that the comment for DataBlockUnitSize.UNSPECIFICIED says the default size is BYTES. If so, wouldn't it make sense to have BYTES be the default if the user doesn't specify a unit?

michaelvdow: That's a good point. If a user doesn't assign a unit size we could just default to BYTES, and appending BYTES when unparsing would just give the user an equivalent statement

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,33 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+SqlCreateAttribute CreateTableAttributeDataBlockSize() :
+{
+    DataBlockModifier modifier = null;
+    DataBlockUnitSize unitSize;
+    SqlLiteral dataBlockSize = null;
+}
+{
+    (
+        (
+            ( <MINIMUM> | <MIN> ) { modifier = DataBlockModifier.MINIMUM; }
```
dasch-1, line 248: Nice, I like this pattern.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -340,6 +340,60 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String expected = "CREATE TABLE `FOO`, DATABLOCKSIZE = 42.123 KILOBYTES (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeDataBlockSizeValueKilobytes() {
+    final String sql = "create table foo, datablocksize = 2e4 kilobytes (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, DATABLOCKSIZE = 2E4 KILOBYTES (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
```
dstekol, line 396: Could also add a test case checking that having both a keyword before DATABLOCKSIZE and an assignment after (ex "MAX DATABLOCKSIZE = 42" ) fails.

dasch-1: +1

michaelvdow: Good point, I'll add a test case for that

PR #49, opened 2020-06-09
[CALCITE] Added support for CREATE TABLE attribute WITH ISOLATED LOADING

michaelvdow:


babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeIsolatedLoading.java
```diff
@@ -0,0 +1,91 @@
...
+ * A <code>SqlCreateAttributeIsolatedLoading</code> is a CREATE TABLE option
+ * for the WITH ISOLATED LOADING attribute.
+ */
+public class SqlCreateAttributeIsolatedLoading extends SqlCreateAttribute {
+
+  private boolean no;
+  private boolean concurrent;
+  private OperationLevel operationLevel;
+
+
```
AndrewPochapsky, line 32: nit: Remove extra new line.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeIsolatedLoading.java
```diff
@@ -0,0 +1,90 @@
...
+public class SqlCreateAttributeIsolatedLoading extends SqlCreateAttribute {
+
+  private boolean no;
+  private boolean concurrent;
+  private OperationLevel operationLevel;
+
+  /**
+   * Creates a {@code SqlCreateAttributeIsolatedLoading}.
+   *
+   * @param no  Defines table as non-load isolated
```
dstekol, line 35: I would recommend something a little more descriptive than "no", maybe "nonLoadIsolated"

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -340,6 +340,56 @@ private void checkParseInfixCast(String sqlType) {
...
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testTableAttributeIsolatedLoadingModifierOperationLevel() {
+    final String sql = "create table foo, with concurrent isolated loading for none (bar integer)";
+    final String expected = "CREATE TABLE `FOO`, WITH CONCURRENT ISOLATED LOADING FOR NONE "
+        + "(`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
```
dstekol, line 392: Combining all the options with a "WITH NO CONCURRENT ISOLATED LOADING FOR ALL" test could be good too. Might be overkill, up to you.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeIsolatedLoading.java
michaelvdow: Yeah no is pretty unclear, I'll change the name

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
michaelvdow: I think the last test there might be enough since we have "no" and "concurrent" already together in a test, and another test with a modifier and the "for" keyword together

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateAttributeIsolatedLoading.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec, final int rightPrec) {
+    writer.keyword("WITH");
+    if (nonLoadIsolated) {
+      writer.keyword("NO");
+    }
+    if (concurrent) {
+      writer.keyword("CONCURRENT");
+    }
+    writer.keyword("ISOLATED");
```
ArthurInTheShell, line 56: nit: Multiple keywords can be chained together

PR #51, opened 2020-06-09
[CALCITE] Add support for AT LOCAL and AT TIME ZONE

ArthurInTheShell:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -441,3 +441,32 @@ SqlNode InlineModOperator() :
         return createCall(qualifiedName, s.end(this), funcType, quantifier, args);
     }
 }
+
+SqlNode dateTimeTerm() :
```
dasch-1, line 445: nit: To be consistent with the style used in the parser, function names should start with uppercase.

babel/src/main/java/org/apache/calcite/sql/babel/SqlDateTimeAtTimeZone.java
```diff
@@ -0,0 +1,65 @@
...
+    return OPERATOR;
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(dateTimePrimary);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    dateTimePrimary.unparse(writer, leftPrec, rightPrec);
+    writer.keyword("AT");
```
dasch-1, line 57: nit: You can actually chain multiple keywords together, which could make things a little more readable. For example:
https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/SqlMerge.java#L190

babel/src/main/java/org/apache/calcite/sql/babel/SqlDateTimeAtLocal.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(dateTimePrimary);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    dateTimePrimary.unparse(writer, leftPrec, rightPrec);
+    writer.keyword("AT");
```
dasch-1, line 53: nit: You can actually chain multiple keywords together, which could make things a little more readable. For example:
https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/SqlMerge.java#L190

babel/src/main/java/org/apache/calcite/sql/babel/SqlDateTimeAtLocal.java
```diff
@@ -0,0 +1,61 @@
...
+      new SqlSpecialOperator("AT LOCAL", SqlKind.OTHER);
+
+  private final SqlNode dateTimePrimary;
+
+  public SqlDateTimeAtLocal(SqlParserPos pos, SqlNode dateTimePrimary) {
+    super(pos);
+    this.dateTimePrimary = dateTimePrimary;
+  }
+
+
```
michaelvdow, line 43: nit: Remove extra newline

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -431,6 +431,60 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testDateTimePrimaryColumn() {
+    final String sql = "select time_column";
```
dasch-1, line 435: nit: Suggest choosing a name like "foo" or "abc" to quickly indicate to the reader that the name is arbitrary.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -431,6 +431,60 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test public void testDateTimePrimaryLiteralAtLocal() {
+    final String sql = "select timestamp '2020-05-30 13:20:00' at local";
+    final String expected = "SELECT TIMESTAMP '2020-05-30 13:20:00' AT LOCAL";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testDateTimePrimaryColumnAtLocal() {
+    final String sql = "select time_column at local";
```
dasch-1, line 459: nit: Same advice here about the column name.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -431,6 +431,60 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test public void testDateTimePrimaryLiteralAtTimeZone() {
+    final String sql = "select timestamp '2020-05-30 13:20:00' at time zone \"GMT+10\"";
+    final String expected = "SELECT TIMESTAMP '2020-05-30 13:20:00' AT TIME ZONE `GMT+10`";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testDateTimePrimaryColumnAtTimeZone() {
+    final String sql = "select time_column at time zone \"Europe Moscow\"";
```
dasch-1, line 477: nit: Same advice here about the column name.

babel/src/main/java/org/apache/calcite/sql/babel/SqlDateTimeAtLocal.java
ArthurInTheShell: Thanks for reminding me. Will fix now

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Change is made.

babel/src/main/java/org/apache/calcite/sql/babel/SqlDateTimeAtTimeZone.java
ArthurInTheShell: Change is made.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Change is made.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -431,6 +431,60 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String expected = "SELECT `TIME_COLUMN` AT TIME ZONE `Europe Moscow`";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testDateTimePrimaryBuiltInFunctionAtTimeZone() {
+    final String sql = "select current_date at time zone \"GMT-5\"";
+    final String expected = "SELECT CURRENT_DATE AT TIME ZONE `GMT-5`";
+    sql(sql).ok(expected);
+  }
+
```
michaelvdow, line 487: I noticed in the design docs that it says this syntax can also appear elsewhere besides a "select" query. It might be worthwhile to add a test where this syntax is in a different type of query to show it works outside of a select statement

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -431,6 +431,60 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testDateTimePrimaryColumn() {
```
dasch-1, line 434: I think we can remove this test case because it doesn't really have to do with this change and other existing tests should break if generic identifier parsing were broken by this change.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Excellent suggestion! Will add more test case to show that this would work in a where clause.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Make sense. What do you feel about the testDateTimePrimaryLiteral and the testDateTimePrimaryBuiltInFunction test case? These grammars are supported by the base parser too, maybe I should remove them too?

dasch-1: Up to you. That's existing functionality and so should already be covered by existing tests. However, at least those are closely related to the change you're making, so I could see keeping them here as well.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Changes made in https://github.com/googleinterns/calcite/pull/51/commits/f542d9676b99e435baa4fdf824909c6638afc482

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Changes made in https://github.com/googleinterns/calcite/pull/51/commits/afb4b0caf81b83fa6fec7980e56781d5fb64ec69

babel/src/main/codegen/config.fmpp
```diff
@@ -922,6 +924,11 @@ data: {
       "InlineModOperator()"
     ]
 
+    dateTimeExpressionSuffixes: [
+      "SqlDateTimeAtLocal"
```
dasch-1, line 928: Suggest you remove the existing entries from this list, insert DateTimeTerm, and then use the <#list> syntax in Parser.jj to iterate through all the functions in this list (which will only have one at this point in time).

ArthurInTheShell: Change is made https://github.com/googleinterns/calcite/pull/51/commits/74cf85b23c39b842cf19b77b2430eabfe901c1b4

PR #52, opened 2020-06-09
[CALCITE] Added support for CREATE TABLE attribute FREESPACE

michaelvdow:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,25 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
     { return new SqlCreateAttributeJournalTable(id, getPos()); }
 }
 
+// FREESPACE attribute can take in decimals but should truncate to an integer
```
AndrewPochapsky, line 239: nit: punctuation.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -388,6 +388,36 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String expected = "(?s).*Encountered \"-\" at .*";
+    sql(sql).fails(expected);
+  }
+
+  @Test public void testTableAttributeFreeSpaceOutOfRangeFails() {
+    final String sql = "create table foo, freespace = ^82.5^ (bar integer)";
+    final String expected = "(?s).*Numeric literal.*out of range.*";
+    sql(sql).fails(expected);
+  }
+
```
dstekol, line 420: Could also add failure case for percent option.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -236,6 +236,25 @@ SqlCreateAttribute CreateTableAttributeJournalTable() :
...
+SqlCreateAttribute CreateTableAttributeFreeSpace() :
+{
+    SqlLiteral tempNumeric;
+    int freeSpaceValue;
+    boolean percent = false;
+}
+{
+    <FREESPACE> <EQ> tempNumeric = UnsignedNumericLiteral() {
+        freeSpaceValue = tempNumeric.getValueAs(Integer.class);
+        if (freeSpaceValue < 0 || freeSpaceValue > 75) {
```
ArthurInTheShell, line 249: It would be nice to have edge case tests, for example the value equals 0, -1, 75.

PR #63, opened 2020-06-11
[CALCITE] Refactored existing column-level attributes to make way for CHARACTER SET attribute

AndrewPochapsky:


core/src/main/java/org/apache/calcite/sql/SqlDataTypeSpec.java
```diff
@@ -168,12 +163,8 @@ public Boolean getNullable() {
...
   }
 
-  public Boolean getUppercase() {
-    return uppercase;
-  }
-
-  public Boolean getCaseSpecific() {
-    return caseSpecific;
+  public List<SqlColumnAttribute> getColumnAttributes() {
+    return columnAttributes;
```
dstekol, line 167: Both the uppercase and casespecific classes have their states stored in private fields with no getter method. So whereas before it was possible to directly access the uppercase and caseSpecific boolean fields via the getUppercase and getCaseSpecific methods, now the only way is to call getColumnAttributes(), unparse each SqlColumnAttribute, and then essentially parse it again. This might not be a problem if getUppercase and getCaseSpecific were never really used, but if there is a scenario in which the state of the column attributes needs to be accessed, it might be worth adding some getter methods to the ColumnAttributes.

AndrewPochapsky: The getters before were only for unparsing purposes so I think it's fine to get rid of them for now. If they do need to get accessed at another point we can re-add them.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -692,7 +692,7 @@ private void checkParseInfixCast(String sqlType) {
     final String sql = "create table foo (bar int uppercase null casespecific, "
         + "baz varchar(30) casespecific uppercase null)";
     final String expected = "CREATE TABLE `FOO` (`BAR` INTEGER UPPERCASE "
-        + "CASESPECIFIC, `BAZ` VARCHAR(30) UPPERCASE CASESPECIFIC)";
+        + "CASESPECIFIC, `BAZ` VARCHAR(30) CASESPECIFIC UPPERCASE)";
```
dstekol, line 695: Curious - how come the casespecific and uppercase parts are switched in output here, but not in the next test? Perhaps it would be possible to modify the way ColumnAttributes are added to the list of ColumnAttributes in order to ensure the output order matches the input order?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -147,67 +147,67 @@ boolean IsNullable() :
...
-    Boolean caseSpecific = null;
+    List<SqlColumnAttribute> columnAttributes =
+        new ArrayList<SqlColumnAttribute>();
     final Span s = Span.of();
 }
 {
     id = CompoundIdentifier()
     type = DataType()
-    // This acts as a loop to check which optional parameters have been specified.
+    /* This structure is to support [NOT] NULL appearing anywhere in the
```
michaelvdow, line 199: How come nullable is not also a ColumnAttribute class? Would converting it to one make it easier to parse all the column attributes in a loop rather than having a seperate boolean for null?

core/src/main/java/org/apache/calcite/sql/SqlColumnAttribute.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * A <code>SqlColumnAttribute</code> is a base class that can be used
+ * to create custom attributes for columns created by the the SQL CREATE TABLE
+ * function.
+ *
+ * <p>To customize column attribute unparsing, override the method
```
michaelvdow, line 26: nit: Missing "</p>"

babel/src/main/codegen/includes/parserImpls.ftl
AndrewPochapsky: Nullable is a special case and is referenced in quite a few places. I spoke with Drew and we decided that it was best to just leave it how it was.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
AndrewPochapsky: Before this change the unparsing had a fixed order as the unparsing logic was simply checking if booleans were checked in a fixed order. Now the attributes are added to the list as they are parsed so the order is no longer fixed. I agree that it would be nice to have a fixed order but I don't think that it's worth the increased complexity. Also I'm pretty sure that the input order does match the output order, right?

AndrewPochapsky: I mean the order matches with the exception of NULL as that is treated differently.

dstekol: Oh, you are right, there are two separate declarations here with different orders, so I thought the order switched.

server/src/main/java/org/apache/calcite/sql/ddl/SqlColumnDeclaration.java
```diff
@@ -70,17 +71,14 @@
...
-      writer.keyword(keyword);
-    }
-    if (dataType.getCaseSpecific() != null) {
-      String keyword = dataType.getCaseSpecific()
-          ? "CASESPECIFIC"
-          : "NOT CASESPECIFIC";
-      writer.keyword(keyword);
+    List<SqlColumnAttribute> columnAttributes = dataType.getColumnAttributes();
+    if (columnAttributes != null && columnAttributes.size() > 0) {
+      SqlWriter.Frame frame = writer.startList("", "");
```
dasch-1, line 76: nit: is this needed since the starting and ending tokens are both empty strings?

server/src/main/java/org/apache/calcite/sql/ddl/SqlColumnDeclaration.java
```diff
@@ -70,17 +71,14 @@
...
+      for (SqlColumnAttribute a : columnAttributes) {
+        writer.sep("");
```
dasch-1, line 78: nit: is this needed since the separator is an empty string?

PR #64, opened 2020-06-11
[CALCITE] Added support for alternative type conversion for literals and identifiers

michaelvdow:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -3626,6 +3626,11 @@ SqlNode Expression3(ExprContext exprContext) :
 <#list parser.preAtomicRowMethods as method>
     e = ${method} { return e; }
 |
+</#list>
+<#list parser.alternativeTypeConversionLiteralOrIdentifier as method>
```
dasch-1, line 3630: It seems a little weird to have a list of methods (alternativeTypeConversionLiteralOrIdentifier) that has the same name as the only method in the list (AlternativeTypeConversionLiteralOrIdentifier). It seems like AlternativeTypeConversionLiteralOrIdentifier might be able to fit into a more general list of methods.

Could we instead add it to preAtomicRowMethods? It looks like the difference is the syntactic lookahead being added here. But maybe we could just add that for all preAtomicRowMethods?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -692,3 +692,80 @@ SqlNode DateTimeTerm() :
...
+    final SqlTypeNameSpec typeNameSpec;
+    final SqlIdentifier typeName;
+    final Span s = Span.of();
+}
+{
+    (
+<#-- additional types are included here -->
+<#-- put custom data types in front of Calcite core data types -->
+<#list parser.dataTypeParserMethods as method>
+        LOOKAHEAD(2)
```
dasch-1, line 730: Isn't Babel's default LOOKAHEAD 2? If so, is this (and the one below) needed?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -692,3 +692,80 @@ SqlNode DateTimeTerm() :
...
+        return new SqlDataTypeSpec(
+            typeName,
+            s.end(this));
+    }
+}
+
+// This excludes CompoundIdentifier() as a type name that's found in the
+// original TypeName() function. Custom data types can be parsed
+// in parser.dataTypeParserMethods.
+SqlTypeNameSpec TypeNameAlternativeCastSyntax() :
```
dasch-1, line 719: This looks like it shares a lot of logic in common with the existing TypeName() function in Parser.jj. Would another approach be to change that function to take an argument indicating whether a compound identifier is valid?

Then we can call that function from here with the argument set to false. And the existing places that invoke it can call with that argument set to true.

WDYT?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -692,3 +692,80 @@ SqlNode DateTimeTerm() :
         )
     )
 }
+
+// This excludes CompoundIdentifier() as a data type.
+SqlDataTypeSpec DataTypeAlternativeCastSyntax() :
```
dasch-1, line 697: If we go with the recommendation below, this function can just be removed and the existing DataType() function in Parser.jj can also have a boolean parameter for whether compound identifiers are allowed.

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: This comment is moot if we go with the other suggestion.

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Yeah that would be a good idea to remove some of this copied code. If we had TypeName with a parameter, would we have still copy the logic in DataType into our custom parsing method? Then the DataType method would just call TypeName with true for compound identifiers?

michaelvdow: Or did you mean adding a parameter to DataType() and passing that down to TypeName(), then just updating the other DataType calls?

dasch-1: Exactly.

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Yeah I'm not sure why they had that in the original TypeName method if that's the default for babel, but I agree with the other suggestion to add a parameter

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Cool that sounds like it'll make this a lot cleaner

core/src/main/codegen/templates/Parser.jj
michaelvdow: Yeah I think it can just go in preAtomicRowMethods, and adding the syntactic lookahead shouldn't hurt the other methods in that array

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Actually I'm not sure how to use a parameter like this to modify the parsing logic because I don't think you can call parsing logic like TypeName() inside a java if statement. Is there some other way to reference this variable inside parsing logic?

PR #67, opened 2020-06-12
[CALCITE] Parse CURRENT_DATE (and friends) as a SqlUnresolvedFunction

AndrewPochapsky:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -663,6 +663,39 @@ SqlNode InlineModOperator() :
     }
 }
 
+SqlNode CurrentDateTimeFunctions() :
```
AndrewPochapsky, line 666: I think adding this caused the build issue, anyone have any idea of how it could be fixed?

Luminarys:
I believe all these functions are available in [Calcite](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/fun/SqlCurrentDateFunction.html). If possible we should parse to them.

I think it's not necessary, but you could also consider supporting all functions like this (they're called niladic by Calcite, see [here](https://calcite.apache.org/docs/reference.html).

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -663,6 +663,69 @@ SqlNode InlineModOperator() :
     }
 }
 
+SqlNode CurrentTimeStampFunction() :
```
dasch-1, line 666: nit: lowercase "S" in timestamp (here and elsewhere).

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -663,6 +663,69 @@ SqlNode InlineModOperator() :
...
+{
+    final List<SqlNode> args = new ArrayList<SqlNode>();
+    SqlNode e;
+}
+{
+    <CURRENT_TIMESTAMP>
+    [
+        <LPAREN>
+        [
+            e = Expression(ExprContext.ACCEPT_SUB_QUERY) {
```
dasch-1, line 676: Can we use AtomicRowExpression here instead of the more general Expression? Same for other functions as well.

AndrewPochapsky: Actually I don't think we can. It looks like a case like SELECT current_time(x+y) fails when using AtomicRowExpression.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -992,7 +992,7 @@ private void checkParseInfixCast(String sqlType) {
 
   @Test public void testCurrentTimestampFunctionBigQuery() {
     final String sql = "select current_timestamp";
```
dasch-1, line 994: Can we delete all of the BigQuery-specific DATE/TIME tests since they are unparsing without parens now?

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: CURRENT_TIME with a math expression as an input parameter doesn't make a whole lot of sense to me. Can you give a link to what test case this is covering?

AndrewPochapsky: https://github.com/googleinterns/calcite/blob/f9ab6107abe6dee5f8aa27239d41c0717bf43366/core/src/test/java/org/apache/calcite/sql/parser/SqlParserTest.java#L4208

dasch-1: Thanks. That seems like a bizarre test case. All of the docs I've checked don't show support for that syntax.

Nevertheless, since the test already exists, we might as well support it.

PR #71, opened 2020-06-13
[CALCITE] Add support for PRIMARY INDEX

dstekol:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
     {
         return new SqlCreateTable(s.end(this), replace, setType, volatility, ifNotExists, id,
-            tableAttributes, columnList, query, withData, onCommitType);
+            tableAttributes, columnList, query, withData, primaryIndex, onCommitType);
     }
 }
 
+/**
+Parses an index declaration.
+Currently only supports PRIMARY INDEX statements, but can be extended to support non-primary indices
```
AndrewPochapsky, line 528: nit: punctuation.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+{
+   List<SqlNode> columns;
+   SqlIdentifier name = null;
+   boolean isUnique = false;
+}
+{
+   (
+       <NO> <PRIMARY> <INDEX>
+       {
+           return new SqlPrimaryIndex(s.end(this), /*columns*/ null, /*name*/ null, /*isUnique*/ false, /*explicitNoPrimaryIndex*/ true);
```
AndrewPochapsky, line 540: nit: for the comments specifying null being passed in use "=" as well. So instead of /\*columns\*/ do /\*columns=\*/

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateTable.java
```diff
@@ -53,20 +54,29 @@
   public SqlCreateTable(SqlParserPos pos, boolean replace, SetType setType, Volatility volatility,
       boolean ifNotExists, SqlIdentifier name, SqlNodeList columnList, SqlNode query) {
     this(pos, replace, setType, volatility, ifNotExists, name, /*tableAttributes=*/null,
-        columnList, query, WithDataType.UNSPECIFIED, OnCommitType.UNSPECIFIED);
+        columnList, query, WithDataType.UNSPECIFIED,
+        /*primaryIndex*/null, OnCommitType.UNSPECIFIED);
```
AndrewPochapsky, line 58: nit: same as above and add space between comment and null.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+       ]
+       <PRIMARY> <INDEX>
+       [
+           name = SimpleIdentifier()
+       ]
+       <LPAREN>
+       columns = SelectList()
+       <RPAREN>
+       {
+           return new SqlPrimaryIndex(s.end(this), columns, name, isUnique, /*explicitNoPrimaryIndex*/ false);
```
AndrewPochapsky, line 554: nit: same as above

babel/src/main/java/org/apache/calcite/sql/babel/SqlIndex.java
```diff
@@ -0,0 +1,49 @@
...
+  public SqlIndex(SqlParserPos pos, List<SqlNode> columns, SqlIdentifier name, boolean isUnique) {
+    this.pos = pos;
+    this.columns = columns;
+    this.name = name;
+    this.isUnique = isUnique;
+  }
+
+  public abstract void unparse(SqlWriter writer, int leftPrec, int rightPrec);
+
+  public SqlParserPos getParserPos() {
```
AndrewPochapsky, line 46: Is this necessary if pos is already public?

babel/src/main/java/org/apache/calcite/sql/babel/SqlPrimaryIndex.java
```diff
@@ -0,0 +1,61 @@
...
+        SqlWriter.Frame frame = writer.startList("(", ")");
+        for (SqlNode c : columns) {
+          writer.sep(",");
+          c.unparse(writer, 0, 0);
+        }
+        writer.endList(frame);
+      }
+    }
+  }
+
```
AndrewPochapsky, line 60: nit: extra new line

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     |
         { query = null; }
     )
+    (
+       index = SqlCreateTableIndex(s)
+       {
+           if(index instanceof SqlPrimaryIndex){
+               primaryIndex = (SqlPrimaryIndex) index;
```
AndrewPochapsky, line 514: Is this casting necessary? Would it work to instead have primary index be of type SqlIndex (in the SqlCreateTable class) and then just pass in the index as is to the constructor? Unparsing should be unaffected.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     |
         { query = null; }
     )
+    (
+       index = SqlCreateTableIndex(s)
+       {
+           if(index instanceof SqlPrimaryIndex){
```
AndrewPochapsky, line 513: nit: include a space between the parentheses: if(foo){} => if (foo) {}

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+}
+{
+   (
+       <NO> <PRIMARY> <INDEX>
+       {
+           return new SqlPrimaryIndex(s.end(this), /*columns*/ null, /*name*/ null, /*isUnique*/ false, /*explicitNoPrimaryIndex*/ true);
+       }
+   |
+       [
+           <UNIQUE> {isUnique = true;}
```
michaelvdow, line 544: nit: space between curly braces and statement

babel/src/main/java/org/apache/calcite/sql/babel/SqlIndex.java
michaelvdow: Maybe the above variables could be made private and have getters to access them in the child class

babel/src/main/java/org/apache/calcite/sql/babel/SqlPrimaryIndex.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import java.util.List;
+
+/**
+ * A <code>SqlPrimaryIndex</code> is a class that can be used
+ * to create a primary index, which is used by the SQL CREATE TABLE function.
+ */
+public class SqlPrimaryIndex extends SqlIndex {
+  boolean explicitNoPrimaryIndex;
```
michaelvdow, line 31: This could probably be made private and final

babel/src/main/java/org/apache/calcite/sql/babel/SqlIndex.java
dstekol: I can make everything protected (I think making them private would make things difficult in the child class). Not sure we'd need getters for all of them since it seems pretty common to make fields only accessible via unparse()
The getParserPos() method is mainly there because it exists in the SqlNode class, and I was trying to conform to that.

babel/src/main/codegen/includes/parserImpls.ftl
dstekol: The idea is that we can eventually extend the SqlIndex class to include other types of indices, not just primary indices. However, we wouldn't want any of those other SqlIndex subclasses to be passed in to the primaryIndex field of SqlCreateTable, since that should only ever be a SqlPrimaryIndex.

AndrewPochapsky: Ah alright, makes sense.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -474,6 +474,8 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     final SqlNodeList columnList;
     final SqlNode query;
     WithDataType withData = WithDataType.UNSPECIFIED;
+    SqlPrimaryIndex primaryIndex = null;
+    SqlIndex index;
```
dasch-1, line 478: nit: Can this be marked as final?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
     |
         { query = null; }
     )
+    (
+       index = SqlCreateTableIndex(s)
+       {
+           if (index instanceof SqlPrimaryIndex) {
+               primaryIndex = (SqlPrimaryIndex) index;
+           }
+           //code for non-primary index support can be added here
```
dasch-1, line 516: nit: We can probably remove this comment as I think the code makes this pretty clear.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+{
+   List<SqlNode> columns;
+   SqlIdentifier name = null;
+   boolean isUnique = false;
+}
+{
+   (
+       <NO> <PRIMARY> <INDEX>
+       {
+           return new SqlPrimaryIndex(s.end(this), /*columns=*/ null, /*name=*/ null, /*isUnique=*/ false, /*explicitNoPrimaryIndex=*/ true);
```
dasch-1, line 540: nit: Line wrap

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
     onCommitType = OnCommitTypeOpt()
     {
         return new SqlCreateTable(s.end(this), replace, setType, volatility, ifNotExists, id,
-            tableAttributes, columnList, query, withData, onCommitType);
+            tableAttributes, columnList, query, withData, primaryIndex, onCommitType);
     }
 }
 
+/**
+Parses an index declaration.
```
dasch-1, line 527: nit: indentation

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+Currently only supports PRIMARY INDEX statements, but can be extended to support non-primary indices.
```
dasch-1, line 528: nit: Line wrap

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -505,13 +507,56 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+   |
+       [
+           <UNIQUE> { isUnique = true; }
+       ]
+       <PRIMARY> <INDEX>
+       [
+           name = SimpleIdentifier()
+       ]
+       <LPAREN>
+       columns = SelectList()
```
dasch-1, line 551: Is SelectList() the right parsing function to call here? According to its doc it: "Parses the select list of a SELECT statement."

babel/src/main/java/org/apache/calcite/sql/babel/SqlPrimaryIndex.java
```diff
@@ -0,0 +1,60 @@
...
+  public SqlPrimaryIndex(SqlParserPos pos, List<SqlNode> columns,
+                         SqlIdentifier name, boolean isUnique, boolean explicitNoPrimaryIndex) {
+    super(pos, columns, name, isUnique);
+    this.explicitNoPrimaryIndex = explicitNoPrimaryIndex;
+  }
+
+  public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    if (explicitNoPrimaryIndex) {
+      writer.keyword("NO PRIMARY INDEX");
+    } else {
```
dasch-1, line 42: nit: Add a return in the if clause above. Then this entire section below can be unnested by one level (which helps with readability).

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: Maybe this entire section (including the left and right parens) could be replaced with
ParenthesizedSimpleIdentifierList?

babel/src/main/java/org/apache/calcite/sql/babel/SqlIndex.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import java.util.List;
+
+/**
+ * A <code>SqlIndex</code> is a class that can be used
```
dasch-1, line 27: nit: This linewrapping seems a bit too aggressive. Suggest wrapping after 80 characters.

babel/src/main/java/org/apache/calcite/sql/babel/SqlPrimaryIndex.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import java.util.List;
+
+/**
+ * A <code>SqlPrimaryIndex</code> is a class that can be used
```
dasch-1, line 27: nit: Same advice on line wrapping here.

babel/src/main/codegen/includes/parserImpls.ftl
dstekol: Thanks, I was looking for something like that and SelectList() seemed close enough, but I will keep this one in mind now.

babel/src/main/codegen/includes/parserImpls.ftl
dstekol: Actually, I realized it can't, because if multiple PRIMARY INDEX statements are declared, the way we defined behavior is that each statement will overwrite the previous one (in order to ensure that there is only ever one primary index). So for example:
CREATE TABLE FOO PRIMARY INDEX (fname) PRIMARY INDEX (lname)
is legal (but only the last one will persist).

dasch-1: Ah you're right. I missed that this was in an ()* clause.

PR #73, opened 2020-06-15
[CALCITE] Added support for QUALIFY clause in SELECT statement

michaelvdow:


babel/src/main/codegen/config.fmpp
```diff
@@ -1001,6 +1002,10 @@ data: {
       "DateTimeTerm()"
     ]
 
+    qualifyClauseMethod: [
```
AndrewPochapsky, line 1005: nit: Should there be an 's' at the end of this? Not sure if we'll have multiple <QUALIFY> statements but even if not adding an 's' keeps with the convention.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1312,20 +1313,26 @@ SqlSelect SqlSelect() :
         where = WhereOpt()
         groupBy = GroupByOpt()
         having = HavingOpt()
+<#list parser.qualifyClauseMethod as method>
+        qualify = ${method}
```
AndrewPochapsky, line 1317: nit: Should there be a \<#sep\> here?

michaelvdow: I'm not really sure how <#sep> could be used here. Could you clarify?

michaelvdow: Also maybe this could be clearer if qualifyClauseMethod was just a boolean and this was an if statement since there should only be one method inside this array

AndrewPochapsky: Don't we usually add \<#sep\> | \</#sep\> when we inject a list of methods?

michaelvdow: I think that's only if each list item is an "or" statement. [<#sep>](https://freemarker.apache.org/docs/ref_directive_list.html#ref_list_sep) is used to display something between each item but not after the last item, but I we don't need "|" inserted here

dasch-1: If there are multiple qualifyClauseMethod's, wouldn't we want a "|" between them? If there is just meant to be a single qualify parse method, maybe we should convert this to a #if instead?

michaelvdow: I think since there should only be one qualify parse method, it would be better to just make this an #if statement. Otherwise #list is kind of misleading in this case

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -171,6 +190,14 @@ public void setHaving(SqlNode having) {
     this.having = having;
   }
 
+  public final SqlNode getQualify() {
+    return qualify;
+  }
+
+  public void setQualify(SqlNode qualify) {
```
dstekol, line 197: Is there a situation where the qualify field would be changed after the SqlSelect object is constructed?

core/src/main/java/org/apache/calcite/sql/SqlSelectOperator.java
```diff
@@ -69,11 +70,12 @@ public SqlCall createCall(
...
         operands[5],
-        (SqlNodeList) operands[6],
+        operands[6],
         (SqlNodeList) operands[7],
-        operands[8],
+        (SqlNodeList) operands[8],
         operands[9],
-        (SqlNodeList) operands[10]);
+        operands[10],
+        (SqlNodeList) operands[11]);
```
dstekol, line 78: It might be good to add a comment to each line mentioning which field each operand corresponds to (ex. that operands[0] corresponds to keywordList, etc).

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
michaelvdow: I'm not sure, I added it since the other parameters also have getters and setters. But if it's not needed I can remove it

core/src/main/java/org/apache/calcite/sql/SqlSelectOperator.java
michaelvdow: Sure I can add some comments here

core/src/main/codegen/config.fmpp
```diff
@@ -452,6 +452,9 @@ data: {
     dateTimeExpressionMethods: [
     ]
 
+    qualifyClauseMethod: [
```
dasch-1, line 455: Should we delete this (here and other config.fmpp files)?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1312,20 +1313,26 @@ SqlSelect SqlSelect() :
         where = WhereOpt()
         groupBy = GroupByOpt()
         having = HavingOpt()
+<#if parser.includeQualifyClause>
+        qualify = QualifyOpt()
+<#else>
+        { qualify = null; }
```
dasch-1, line 1319: To simplify the parsing logic a bit, can we assign qualify to null in the constructor and then remove this whole else clause?

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
dasch-1: Agree with Daniel. We can remove this setter if there aren't any users today to make it clear that it should be initialized in the constructor.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -44,6 +44,7 @@
   SqlNode where;
   SqlNodeList groupBy;
   SqlNode having;
+  SqlNode qualify;
```
dasch-1, line 47: Even though this isn't the pattern used today, let's make this new field private final.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
michaelvdow: Cool SGTM

core/src/main/codegen/config.fmpp
michaelvdow: Yeah I missed that when refactoring, will remove

PR #74, opened 2020-06-15
[CALCITE] Add support for multiple sources tables in UPDATE FROM statement

AndrewPochapsky:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -833,6 +833,47 @@ private void checkParseInfixCast(String sqlType) {
...
+  @Test public void testUpdateFromMultipleSourcesAllAliased() {
+    final String sql = "UPDATE foo as f from bar as b, baz as z SET "
+        + "f.x = b.x, f.y = z.y";
+    final String expected = "UPDATE `FOO` AS `F` FROM `BAR` AS `B`, `BAZ` AS `Z` SET "
+        + "`F`.`X` = `B`.`X`, `F`.`Y` = `Z`.`Y`";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testUpdateFromMultipleSourcesAllAliasedBigQuery() {
+    final String sql = "UPDATE foo AS f FROM bar AS b, baz AS z SET "
```
dasch-1, line 860: nit: we can probably omit this test since the one above covers the BigQuery case.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1698,10 +1700,38 @@ SqlNode SqlUpdate() :
...
     [ [ <AS> ] alias = SimpleIdentifier() ]
 <#if parser.allowUpdateFromTable>
-    (
-        <FROM> secondTable = TableRef()
-        [ [ <AS> ] secondAlias = SimpleIdentifier() ]
-    )?
+    [
+        (
+            <FROM>
+            sourceTable = TableRef() {
```
dasch-1, line 1706: Since this logic is duplicated in the <COMMA> case below, can we break this out into its own parser helper function?

AndrewPochapsky: It looks like it is possible, though I think then the two lists will need to be initialized to be empty before hand right?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1702,34 +1700,15 @@ SqlNode SqlUpdate() :
 <#if parser.allowUpdateFromTable>
     [
         (
-            <FROM>
-            sourceTable = TableRef() {
+             <FROM> {
```
dasch-1, line 1703: nit: indentation

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1666,8 +1666,8 @@ SqlNode SqlDelete() :
 SqlNode SqlUpdate() :
 {
     SqlNode table;
-    SqlNode secondTable = null;
-    SqlIdentifier secondAlias = null;
+    SqlNodeList sourceTables = null;
```
ArthurInTheShell, line 1669: How is this approach compares with having sourceTables initialized as an empty SqlNodeList with the final keyword? Same question for the sourceAliases.

AndrewPochapsky: I don't think it matters much as the only thing that changes is which condition you check for when unparsing, either list != null or list.size() > 0. I suppose making it an empty list is better as it avoids null but it seems like a marginal difference.

ArthurInTheShell: Yeah, it would be up to you

PR #75, opened 2020-06-16
[CALCITE] Parse DATE as a function

AndrewPochapsky:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -35,10 +35,15 @@ SqlNode DateFunctionCall() :
...
     }
-    args = FunctionParameterList(ExprContext.ACCEPT_SUB_QUERY) {
-        quantifier = (SqlLiteral) args.get(0);
-        args.remove(0);
-        return createCall(qualifiedName, s.end(this), funcType, quantifier, args);
+    [
+        args = FunctionParameterList(ExprContext.ACCEPT_SUB_QUERY) {
+            quantifier = (SqlLiteral) args.get(0);
+            args.remove(0);
+            return createCall(qualifiedName, s.end(this), funcType, quantifier, args);
```
AndrewPochapsky, line 42: I tried converting this to use `return SqlStdOperatorTable.DATE.createCall` but I couldn't get it to work properly. For some reason a testcase broke when I removed the `args.remove(0)` line, even if the quantifier wasn't used anywhere (ie I instead returned `SqlStdOperatorTable.DATE.createCall(getPos(), args);`.

AndrewPochapsky: However it seems to be being parsed a function regardless since I had to remove the backticks for a previous test.

AndrewPochapsky: It may be possible to just keep the line `args.remove(0)` and return the new function (that works for the current tests) but I'm not sure if it will work for other cases. For the cases that I've looked at `quantifier` is null.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -841,6 +846,8 @@ SqlNode DateTimeTerm() :
         e = DateTimeLiteral()
     |
         e = SimpleIdentifier()
+    |
+        e = DateFunctionCall()
```
michaelvdow, line 850: DateFunctionCall is also in the [BuiltinFunctionCalls](https://github.com/googleinterns/calcite/blob/master/babel/src/main/codegen/config.fmpp#L963). Does it need to be called in both of these places?

AndrewPochapsky: It does. This one is to allow for DATE AT LOCAL queries and the one in BuiltinFunctionCalls is for standalone DATE queries.

michaelvdow: Ah ok, how come the first call to DateFunctionCall() wouldn't parse both of these?

AndrewPochapsky: I figured that it would be simplest to just use the existing parsing logic for AT LOCAL rather than add more logic (for a fairly edge case too) to DateFunctionCall().

michaelvdow: Oh I see, that makes sense

core/src/main/java/org/apache/calcite/sql/fun/SqlDateFunction.java
```diff
@@ -0,0 +1,58 @@
...
+        SqlKind.OTHER_FUNCTION,
+        ReturnTypes.DATE,
+        null,
+        OperandTypes.NILADIC,
+        SqlFunctionCategory.TIMEDATE);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  public SqlSyntax getSyntax() {
```
dasch-1, line 46: nit: Does this need an @Override tag?

core/src/main/java/org/apache/calcite/sql/fun/SqlDateFunction.java
```diff
@@ -0,0 +1,58 @@
...
+    return SqlSyntax.FUNCTION_ID;
+  }
+
+  @Override public SqlMonotonicity getMonotonicity(SqlOperatorBinding call) {
+    return SqlMonotonicity.INCREASING;
+  }
+
+  // Plans referencing context variables should never be cached
+  public boolean isDynamicFunction() {
```
dasch-1, line 55: Can we just omit this function altogether and let it use the base class version?

AndrewPochapsky: I'm not too sure the significance of this but it seems that it is false by default.

AndrewPochapsky: I have it here in the first place as I was referencing [SqlCurrentDateFunction](https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/fun/SqlCurrentDateFunction.java)

dasch-1: Ok sounds good. Either way seems fine.

PR #76, opened 2020-06-16
[CALCITE] Added support for alternative type syntax for queries

michaelvdow:


babel/src/main/codegen/config.fmpp
```diff
@@ -1027,6 +1027,10 @@ data: {
       "DateTimeTerm()"
     ]
 
+    alternativeTypeConversionQueryMethods: [
```
ArthurInTheShell, line 1030: With this name, are you expecting there will be more entries in "alternativeTypeConversionQueryMethods" in the future?

michaelvdow: No this should be the only method for this specific syntax. Could make it clearer if I convert this to a boolean

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1273,4 +1273,36 @@ private void checkParseInfixCast(String sqlType) {
...
+  @Test void testAlternativeTypeConversionQueryInterval() {
+    final String sql = "select (select foo from bar) (interval minute) from baz";
+    final String expected = "SELECT CAST((SELECT `FOO`\n"
+        + "FROM `BAR`) AS INTERVAL MINUTE)\n"
+        + "FROM `BAZ`";
+    sql(sql).ok(expected);
+  }
+
+  @Test void testAlternativeTypeConversionQueryNested() {
+    final String sql = "select ((select foo from bar) (integer)) (char) from baz";
```
dasch-1, line 1302: nit: Can we change this nested test case such that the (integer) cast is after "foo"? That would be a more realistic scenario I think.

michaelvdow: Sure, updated

PR #77, opened 2020-06-16
[CALCITE] Add support for RENAME TABLE

AndrewPochapsky:


babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+
+/**
+ * Parse tree for {@code RENAME} statement.
+ */
+public class SqlRenameTable extends SqlCall implements SqlExecutableStatement {
+  public static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("RENAME", SqlKind.OTHER);
```
michaelvdow, line 38: Should this be "RENAME TABLE" since there can be other RENAME functions? This would be more in line with the create table special operator

AndrewPochapsky: Yep, will make the change.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -892,6 +892,27 @@ SqlDataTypeSpec DataTypeAlternativeCastSyntax() :
     }
 }
 
+SqlNode SqlRenameTable() :
```
dasch-1, line 895: nit: Should we return a more specific node type (e.g. SqlRenameTable)?

babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+
+/**
+ * Parse tree for {@code RENAME TABLE} statement.
+ */
+public class SqlRenameTable extends SqlCall implements SqlExecutableStatement {
+  public static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("RENAME TABLE", SqlKind.OTHER);
```
dasch-1, line 38: We haven't been too strict about this in past PRs, but we should get in the habit of declaring a new SqlKind type, rather than just using SqlKind.OTHER for everything.

This can be helpful for downstream consumers.

babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
```diff
@@ -0,0 +1,82 @@
...
+    this.sourceTable = sourceTable;
+    this.renameOption = renameOption;
+  }
+
+  @Override public SqlOperator getOperator() {
+    return OPERATOR;
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(targetTable, sourceTable);
```
dasch-1, line 58: Should we include the renameOption in the operand list?

AndrewPochapsky: Doesn't it have to be a SqlNode though?

babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
AndrewPochapsky: Sure, should it be SqlKind.RENAME for this case?

babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
dasch-1: Yeah it does. Wasn't sure if it would be worth wrapping it in some sort of SqlNumericLiteral. But that would probably be overkill. We can just leave it as is.

babel/src/main/java/org/apache/calcite/sql/babel/SqlRenameTable.java
dasch-1: Probably SqlKind.RENAME_TABLE

PR #79, opened 2020-06-16
Create function sql form

ArthurInTheShell:


babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,171 @@
...
+   * @param returnExpression the expression that is returned
+   * */
+  public SqlCreateFunctionSqlForm(final SqlParserPos pos, final boolean replace,
+      final SqlIdentifier functionName,
+      final SqlIdentifier specificFunctionName, final List<SqlIdentifier> fieldNames,
+      final List<SqlDataTypeSpec> fieldTypes,
+      final SqlDataTypeSpec returnsDataType, final DeterministicType isDeterministic,
+      final ReactToNullInputType canRunOnNullInput, final boolean hasSqlSecurityDefiner,
+      final int typeInt, final SqlNode returnExpression) {
+
```
michaelvdow, line 68: nit: extra new line

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -974,6 +974,185 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String expected = "CREATE FUNCTION `ADD1` (`A` INTEGER) "
+        + "RETURNS INTEGER "
+        + "LANGUAGE SQL "
+        + "DETERMINISTIC "
+        + "COLLATION INVOKER INLINE TYPE 1 "
+        + "RETURN (`A` + 1)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testCreateFunctionSqlFormCalledOnNUllInput() {
```
michaelvdow, line 1066: nit: capitalization in this method and the one below, "NU" in Null

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -974,6 +974,185 @@ private void checkParseInfixCast(String sqlType) {
...
+        "create function foo() "
+            + "returns Integer "
+            + "language sql "
+            + "collation invoker inline type ^-^2 "
+            + "return current_date + 1";
+    final String expected = "(?s).*Encountered \"-\" at .*";
+    sql(sql).fails(expected);
+  }
+
+  @Test public void testCreateFunctionSqlFormhasSecurityDefiner() {
```
michaelvdow, line 1139: nit "Formhas" -> "FormHas"

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -565,6 +565,111 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+}
+
+
+/**
+* Parse a "name1 type1 , name2 type2 ..." list,
+* the field type default is not nullable.
+*/
+void FieldNameTypeCommaListWithoutOptionalNull(
+        List<SqlIdentifier> fieldNames,
+        List<SqlDataTypeSpec> fieldTypes) :
```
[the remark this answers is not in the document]
ArthurInTheShell: I believe SqlIdentifier is a specific type of SqlNode, and SqlIdentifier covered the usecases I needed here. Also, this function is more or less a copy of [FieldNameTypeCommaList](https://github.com/googleinterns/calcite/blob/05376d6bff11ccef690addb4f9c40ab16dd20bec/core/src/main/codegen/templates/Parser.jj#L5082) (excluded the optional null token). The field name could either be a simpleIdentifier or a compoundIdentifier. Could you give more reasoning as to why a SqlNode type would be more suitable here?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -565,6 +565,111 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+    SqlLiteral tempNumeric;
+    int typeInt;
+    final SqlNode returnExpression;
+}
+{
+    <FUNCTION>
+    functionName = CompoundIdentifier()
+    <LPAREN>
+    [
+      FieldNameTypeCommaListWithoutOptionalNull(fieldNames, fieldTypes)
```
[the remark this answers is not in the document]
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/79/commits/57dc0c69a50e5b504767b9d71af8b877377d8cd9

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -565,6 +565,111 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+    }
+    <RETURN> returnExpression = Expression(ExprContext.ACCEPT_SUB_QUERY)
+    {
+        return new SqlCreateFunctionSqlForm(s.end(this), replace,
+            functionName, specificFunctionName, fieldNames, fieldTypes,
+            returnsDataType, isDeterministic, canRunOnNullInput,
+            hasSqlSecurityDefiner, typeInt, returnExpression);
+    }
+
+
```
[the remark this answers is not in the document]
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/79/commits/57dc0c69a50e5b504767b9d71af8b877377d8cd9

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -565,6 +565,111 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
...
+}
+
```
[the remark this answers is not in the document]
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/79/commits/57dc0c69a50e5b504767b9d71af8b877377d8cd9

babel/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
AndrewPochapsky: I'm not saying change the type to SqlNode but to to use the SqlNodeList class instead of List<SqlIdentifer> and List<SqlDataTypeSpec>. As you said SqlIdentifier is already a SqlNode so it should work. Though if you are just copying an existing function it's probably best to keep it as is.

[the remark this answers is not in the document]
ArthurInTheShell: I see, you mean the data type SqlNodeList would be useful here compares to the List of SqlIdentifier here. That is a good point. Since this is a modified copy of an existing function, I'll refrain from making more changes than needed unless there is more people prefer SqlNodeList than a list of SqlIdentifier here.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,171 @@
...
+    writer.keyword("COLLATION INVOKER INLINE TYPE");
+    writer.print(typeInt + " ");
+    writer.keyword("RETURN");
+    returnExpression.unparse(writer, 0, 0);
+  }
+
+  public enum DeterministicType {
+    /**
+     * Not Specified
```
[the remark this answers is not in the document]
ArthurInTheShell: Thanks for pointing this out. Fixed in https://github.com/googleinterns/calcite/pull/79/commits/bb4a422f923beb574a38980513239ea72534a49e

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+
+public class SqlCreateFunctionSqlForm extends SqlCreate {
+  public final SqlIdentifier functionName;
+  public final SqlIdentifier specificFunctionName;
+  public List<SqlIdentifier> fieldNames;
+  public List<SqlDataTypeSpec> fieldTypes;
```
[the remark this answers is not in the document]
ArthurInTheShell: Fixes made https://github.com/googleinterns/calcite/pull/79/commits/bb4a422f923beb574a38980513239ea72534a49e

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
ArthurInTheShell: I keep this new line here to improve the readability of the code. The constructor takes in a large amount of argument, therefore it would look like a cluster with the code in the constructor without this new line. I am leaning toward keeping it. Please let me know your opinion on it.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Fixes made in https://github.com/googleinterns/calcite/pull/79/commits/973cb3f514eabcc9dcb2e51bebf64c3d23d9e7c0 . Thank you for pointing it out.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Fixes made in https://github.com/googleinterns/calcite/pull/79/commits/973cb3f514eabcc9dcb2e51bebf64c3d23d9e7c0 .

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
michaelvdow: Yeah I see what you mean. I would probably keep it in that case

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1153,6 +1153,27 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testCreateFunctionSqlFormNotDeterministicReturnsNullHasSecurityDefiner() {
+    final String sql =
+        "create function foo() "
+            + "returns Integer "
+            + "language sql "
+            + "returns null on null input "
+            + "not deterministic "
```
ArthurInTheShell, line 1162: This controversial test case is added. Where "returns null on null input" and "not deterministic" came in in one order but unparsed into another order. It would be grammatically the same query, since the order of "[not] deterministic", "[return null|called] on null input", and "specific <function_name>" is arbitrary. After all, it did unparse into a different query with equivalent meaning, should we be concerned about this behavior? and if so, any suggestions on how to unparse in the order of query input?

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,172 @@
...
+  public final DeterministicType isDeterministic;
+  public final ReactToNullInputType canRunOnNullInput;
+  public final boolean hasSqlSecurityDefiner;
+  public final int typeInt;
+  public final SqlNode returnExpression;
+
+  private static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("CREATE FUNCTION", SqlKind.CREATE_FUNCTION);
+
+  /** Creates a SqlCreateFunctionSqlForm.
```
dasch-1, line 46: nit: The first comment line should be on the line after the "/**".

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,172 @@
...
+   * @param fieldNames function parameter names
+   * @param fieldTypes function parameter types
+   * @param returnsDataType return type of the function
+   * @param isDeterministic if "deterministic" is specified
+   * @param canRunOnNullInput if "called on null input" or "returns null on null input"
+   *                          is specified
+   * @param hasSqlSecurityDefiner if "sql security definer" is specified
+   * @param typeInt integer value after inline type
+   * @param returnExpression the expression that is returned
+   * */
```
dasch-1, line 60: nit: You have an extra "* " prefix here.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -0,0 +1,172 @@
...
+    this.returnsDataType = returnsDataType;
+    this.isDeterministic = isDeterministic;
+    this.canRunOnNullInput = canRunOnNullInput;
+    this.hasSqlSecurityDefiner = hasSqlSecurityDefiner;
+    this.typeInt = typeInt;
+    this.returnExpression = returnExpression;
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(functionName, specificFunctionName, returnsDataType,
```
dasch-1, line 83: Why only return a subset of the fields in the operands list?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dasch-1: No, as long as the behavior is the same, the order shouldn't matter.

ArthurInTheShell: Okay. Thanks for clarifying.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
ArthurInTheShell: I think the fix would be adding FieldNames and FieldTypes to the OperandList, is there more to be added?

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
ArthurInTheShell: Fix made in https://github.com/googleinterns/calcite/pull/79/commits/0ae2a1dacf8f672feaaae88448156fbd8470c66f .

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
ArthurInTheShell: Fix made in https://github.com/googleinterns/calcite/pull/79/commits/0ae2a1dacf8f672feaaae88448156fbd8470c66f .

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
dasch-1: I'm not quite sure what the recommendation is regarding getOperandList. It seems to be primarily referenced in SqlCall's clone() method.

At the very least, we should be declaring a getOperator() override if we have getOperandList() declared.

ArthurInTheShell: getOperator is added https://github.com/googleinterns/calcite/pull/79/commits/a27c1db25901e8d949eb7c29e0ab2057089146d1.

ArthurInTheShell: After looking at the documentation of [SqlOperator](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/SqlOperator.html) and some [other implementation of class implemented SqlCreate](https://github.com/googleinterns/calcite/blob/master/babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateTable.java), I believe it would be reasonable to add FieldNames and FieldTypes to Operand in the format of SqlNodeList.

ArthurInTheShell: Which means I would take [Andrew's advice](https://github.com/googleinterns/calcite/pull/79#discussion_r441524252) and make FieldNames and FieldTypes into the type SqlNodeList, which can be added directly into Operand list.

ArthurInTheShell: Fixes made in https://github.com/googleinterns/calcite/pull/79/commits/55b082e08f165ae803e9ebfc6c1bfa2a29e2b37d.

babel/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
ArthurInTheShell: Updated to use SqlNodeList in https://github.com/googleinterns/calcite/pull/79/commits/55b082e08f165ae803e9ebfc6c1bfa2a29e2b37d.

PR #82, opened 2020-06-17
[CALCITE] Add support for DEFAULT column attribute

AndrewPochapsky:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -171,6 +171,34 @@ boolean IsNullable() :
...
+{
+    <DEFAULT_>
+    (
+        defaultValue = OptionValue()
+    |
+        <NULL> {
+            defaultValue = SqlLiteral.createNull(getPos());
+        }
+    |
+        // The DateTime functions are singled out to allow for arguments to
```
dasch-1, line 187: nit: Suggest moving this to a function-level comment since it applies to multiple branches in this parse clause.

AndrewPochapsky: Sure, you mean moving it right above the declaration right?

dasch-1: Yeah, just above "SqlColumnAttribute ColumnAttributeDefault() :"

PR #86, opened 2020-06-20
[CALCITE] Replace view feature

dstekol:
A lot of the changes were simply to make it possible to have the "replace" field in SqlCreate be an enum instead of a boolean.


core/src/main/java/org/apache/calcite/sql/SqlCreateOrReplace.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+/**
+ *
+ */
+package org.apache.calcite.sql;
+
+/**
+ * Enum used by {@link org.apache.calcite.sql.SqlCreate} to indicate whether the query is a
```
dasch-1, line 24: nit: wrap at 80 characters

core/src/main/java/org/apache/calcite/sql/SqlCreateOrReplace.java
```diff
@@ -0,0 +1,45 @@
...
+ * "CREATE", "CREATE OR REPLACE", or "REPLACE" query
+ */
+public enum SqlCreateOrReplace {
```
dasch-1, line 27: Can we define this enum within the scope of the existing SqlCreate class? That seems to be a pattern we're following and makes organizational sense to me.

dasch-1, line 27: I think "CreateOrReplace" is too ambiguous of a name for this enum. It can be misinterpreted as having either of the two meanings:
- "CREATE OR REPLACE"
- "CREATE" OR "REPLACE"

As such, I think we should steer away from that ambiguity and select something slightly more explicit. How about "enum SqlCreateType" or "enum SqlCreateOption"?

core/src/test/codegen/config.fmpp
```diff
@@ -390,11 +391,16 @@ data: {
     ]
 
     # List of methods for parsing extensions to "CREATE [OR REPLACE]" calls.
-    # Each must accept arguments "(SqlParserPos pos, boolean replace)".
+    # Each must accept arguments "(SqlParserPos pos, SqlCreateOrReplace replace)".
```
dasch-1, line 394: We should change all of the occurrences of "replace" with either "createType" or "createOption" (depending on what your preference is based on my comment on the enum name).

The same is true for all of the locations you've updated throughout the code.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -48,7 +49,7 @@
   /**
    * Creates a SqlCreateFunctionSqlForm.
    * @param pos position
-   * @param replace if "or replace" token occurred
+   * @param replace enum to distinguish between "CREATE", "CREATE OR REPLACE", and "REPLACE"
```
dasch-1, line 52: nit: wrap at 80 characters

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateFunctionSqlForm.java
```diff
@@ -61,7 +62,7 @@
    * @param typeInt integer value after inline type
    * @param returnExpression the expression that is returned
    */
-  public SqlCreateFunctionSqlForm(final SqlParserPos pos, final boolean replace,
+  public SqlCreateFunctionSqlForm(final SqlParserPos pos, final SqlCreateOrReplace replace,
```
dasch-1, line 65: nit: wrap at 80 characters

core/src/main/java/org/apache/calcite/sql/SqlCreateOrReplace.java
```diff
@@ -0,0 +1,45 @@
...
+  CREATE("CREATE"),
+  REPLACE("REPLACE"),
+  CREATE_OR_REPLACE("CREATE OR REPLACE");
+
+  private final String name;
```
dasch-1, line 32: Why use strings at all? Can we just use the raw enum and get rid of the rest of these helper methods in the enum definition?

core/src/test/codegen/config.fmpp
dasch-1, line 394: nit: wrap at 80 characters

core/src/test/codegen/config.fmpp
```diff
@@ -390,11 +391,16 @@ data: {
...
     createStatementParserMethods: [
       "SqlCreateTable"
     ]
 
+    # List of methods for parsing extensions to "REPLACE" calls.
+    # Each must accept arguments "(SqlParserPos pos, SqlCreateOrReplace replace)".
```
dasch-1, line 400: nit: wrap at 80 characters

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,30 @@
...
   }
 
   @Deprecated // to be removed before 2.0
-  public SqlCreate(SqlParserPos pos, boolean replace) {
+  public SqlCreate(SqlParserPos pos, SqlCreateOrReplace replace) {
     this(SqlDdl.DDL_OPERATOR, pos, replace, false);
   }
 
-  public boolean getReplace() {
+  public SqlCreateOrReplace getReplace() {
```
dasch-1, line 47: Let's also change this getter to match the new field name.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,30 @@
...
     return replace;
   }
 
-  public void setReplace(boolean replace) {
```
dasch-1, line 51: Prefer to just delete the setter rather than comment it out.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,30 @@
 public abstract class SqlCreate extends SqlDdl {
 
   /** Whether "OR REPLACE" was specified. */
-  boolean replace;
+  final SqlCreateOrReplace replace;
```
dasch-1, line 29: This field name is a little misleading now that it has a type that can be one of three options. We should rename this field to "createType" or "createOption" (depending on your preference regarding the name of the enum).

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,30 @@
...
 
   /** Whether "IF NOT EXISTS" was specified. */
   public final boolean ifNotExists;
 
   /** Creates a SqlCreate. */
-  public SqlCreate(SqlOperator operator, SqlParserPos pos, boolean replace,
+  public SqlCreate(SqlOperator operator, SqlParserPos pos, SqlCreateOrReplace replace,
```
dasch-1, line 35: Let's also update this parameter name to match the field.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
dstekol: I considered both of those, but they would also be confusing because SqlCreateType and SqlCreateOption both show up elsewhere in the code (presumably corresponding to CREATE TYPE and CREATE OPTION). I also don't love the name SqlCreateOrReplace, definitely open to other suggestions.

ArthurInTheShell: How does "SqlCreateReplace" sound? Also, I would suggest renaming the variable holding this enum value too, since it is no longer a boolean and has either true or false. For the variable, I'll suggest the name of "option" with some prefixes? Such variable name might not be accurate within the project context, but would be very clear within the scope of the class.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,30 @@
...
       boolean ifNotExists) {
     super(operator, pos);
     this.replace = replace;
     this.ifNotExists = ifNotExists;
   }
 
   @Deprecated // to be removed before 2.0
-  public SqlCreate(SqlParserPos pos, boolean replace) {
+  public SqlCreate(SqlParserPos pos, SqlCreateOrReplace replace) {
```
ArthurInTheShell, line 43: When I came across the "@Deprecated" line, I came to realization that this might be a well documented component that the **Adapters** in Calcite project uses. Could you clarify in the design doc how this change won't break codes that depends on this functionality?

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
dasch-1: If CreateType and CreateOption are already taken, how about one of the following for the name of the enum?

- CreateVariant
- CreateModifier
- CreateSpecifier

core/src/main/java/org/apache/calcite/sql/SqlCreateOrReplace.java
dstekol: My idea was to simplify the unparsing logic by making it possible to convert the enum directly to the relevant keywords (rather than creating a switch and writing unparsing code for each case).

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
dstekol: CreateSpecifier sounds good.

core/src/test/codegen/config.fmpp
dstekol: The line is 83 characters, and the text makes sense to group together. Is it really worth it to split the line over 3 extra characters? I feel like that might hurt readability more than it would help.

server/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -410,3 +410,22 @@ SqlDrop SqlDropFunction(Span s, boolean replace) :
...
+
+SqlCreate SqlReplace() :
+{
+    final Span s;
+    final SqlCreate create;
+}
+{
+    <REPLACE> { s = span(); }
+    (
+<#-- additional literal parser methods are included here -->
```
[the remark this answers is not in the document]
dstekol: This code is copied almost exactly from the existing SqlCreate function, so I think it makes sense to follow the same pattern for the sake of consistency.

server/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -410,3 +410,22 @@ SqlDrop SqlDropFunction(Span s, boolean replace) :
         return SqlDdlNodes.dropFunction(s.end(this), ifExists, id);
     }
 }
+
```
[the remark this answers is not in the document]
dstekol: Same response as above - this is done the same way as in the SqlCreate function. I am not totally clear on what the benefit of having SqlDdlNodes here is either, but I think it makes sense to stay consistent with how the existing code handles similar functionalities.

core/src/test/codegen/config.fmpp
dasch-1: 80 characters is a pretty widely-held standard. Even if it may make a line slightly less readable, it's a good idea to stick to that standard in practice. In this case, you could break the line after "arguments" since the remaining portion in quotes is kind of a single unit.

server/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
dasch-1: Sounds good to me. We can always have a follow-up PR to change all locations to be consistent if we decide it would be better.

server/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
dasch-1: I'm not sure I'm following this comment thread.

@ArthurInTheShell do you mean wrap the definition of SqlReplac() in a #if? If so, since it's defined in babel's parserImpls.ftl file, we can omit that.

@dstekol SqlDdlNodes is the name of the class where dropFunction is defined. Can you clarify on what you mean? Is there an alternative to not using its dropFunction method?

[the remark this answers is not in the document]
dstekol: Actually, I think I might be referring to the wrong part of the code. SqlDdlNodes has methods that are used (elsewhere in the code) to create the objects (rather than directly calling the constructor). For example, there is a createView method in SqlDdlNodes that returns a new SqlCreateView object by calling the constructor. I'm not sure why calling this method is better than calling the constructor directly, but this is the pattern that has been used, so I've been trying to stick to it.
@ArthurInTheShell could you clarify what you had in mind?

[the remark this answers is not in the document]
ArthurInTheShell: Drew's understanding of my advice is correct, I think we can omit that since it is in babel/parserImpls.ftl . Btw, I think we should keep the comment as unresolved if it is still open. I saw notification from github saying someone mentioned me in a comment but I can't find where exactly I am being mentioned.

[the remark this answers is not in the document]
dstekol: Ok, sounds good. You're right, I may have been too trigger-happy with the resolve button.

server/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
ArthurInTheShell: It is indeed similar to that of the SqlCreate function in parser.jj, but I am still confuse as to why the LOOKAHEAD(2) is here. Could you briefly explain the reason?

core/src/test/codegen/config.fmpp
dstekol: Ok, will do.

dstekol: Agreed, just changed all "replace" params to createSpecifier.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
dstekol: I'm a little confused - all the parameter types have been updated throughout the project (boolean -> enum), and all the tests are passing. The design doc mentions making this boolean->enum change in all dependencies (I felt there were too many small search-and-replace changes throughout the project for it to be worth listing them all out). Could you clarify why you think this change in particular needs more documentation?

server/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
dstekol: Honestly, I haven't tried modifying it, so I'm not sure what the reasoning is. As I mentioned, I did not make the decision to insert it, just the syntax is very similar to CREATE, so I am trying to stay as true as possible to the existing SqlCreate parsing logic.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -611,12 +611,12 @@ SqlCreate SqlCreateTable(Span s, SqlCreateOrReplace replace) :
     )*
     onCommitType = OnCommitTypeOpt()
     {
-        return new SqlCreateTable(s.end(this), replace, setType, volatility, ifNotExists, id,
+        return new SqlCreateTable(s.end(this), createSpecifier, setType, volatility, ifNotExists, id,
```
dasch-1, line 614: nit: line wrap

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
ArthurInTheShell: I am okay with it since no test fails because of the change.

babel/src/main/codegen/config.fmpp
```diff
@@ -985,12 +985,19 @@ data: {
     ]
 
     # List of methods for parsing extensions to "CREATE [OR REPLACE]" calls.
-    # Each must accept arguments "(SqlParserPos pos, boolean replace)".
+    # Each must accept arguments:
+    # "(SqlParserPos pos, SqlCreate.SqlCreateSpecifier createSpecifier)".
```
ArthurInTheShell, line 989: nit: Add "org.apache.calcite.sql.SqlCreate.SqlCreateSpecifier" in the imports at line 24, and use the Enum name SqlCreateSpecifier rather than SqlCreate.SqlCreateSpecifier in config.fmpp, parserImpls.ftl, and in code.

ArthurInTheShell: Correct me if there is a particular reason that drive you to use SqlCreate.SqlCreateSpecifier rather than SqlCreateSepcifier.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,49 @@
 public abstract class SqlCreate extends SqlDdl {
 
   /** Whether "OR REPLACE" was specified. */
```
ArthurInTheShell, line 28: nit: The comment block need to be change to reflect the change from boolean to enum.

core/src/main/java/org/apache/calcite/sql/SqlCreate.java
```diff
@@ -26,30 +26,49 @@
...
-  boolean replace;
+  final SqlCreateSpecifier createSpecifier;
 
   /** Whether "IF NOT EXISTS" was specified. */
   public final boolean ifNotExists;
 
+  /**
+   * Enum to indicate whether the query is a "CREATE", "CREATE OR REPLACE",
+   * or "REPLACE" query
```
michaelvdow, line 36: nit: missing period

babel/src/main/codegen/config.fmpp
dstekol: Good point, will do.

michaelvdow: I think if it's changed here, it might be good to also change it throughout the Java files because some of them reference SqlCreate.SqlCreateSpecifier and other Java files just reference SqlCreateSpecifier

PR #90, opened 2020-06-23
[CALCITE] Add support for TOP N operator in SELECT statement

dstekol:


babel/src/main/java/org/apache/calcite/sql/babel/SqlSelectTopN.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.List;
+
+/**
+ * <code>SqlSelectTopN</code> is a class that handles the TOP n syntax in
+ * SELECT statements. It is used by the {@link SqlSelect} class.
+ */
+public class SqlSelectTopN extends SqlCall {
+  private static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("TOP", SqlKind.OTHER);
```
AndrewPochapsky, line 39: nit: I believe @dasch-1 said that we should add new elements to SqlKind from now on.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -949,7 +949,16 @@ public void unparseOffsetFetch(SqlWriter writer, SqlNode offset,
    * @param offset Number of rows to skip before emitting, or null
    * @param fetch Number of rows to fetch, or null
    */
-  public void unparseTopN(SqlWriter writer, SqlNode offset, SqlNode fetch) {
+  public void unparseFetchAsTopN(SqlWriter writer, SqlNode offset, SqlNode fetch) {
+    //fetch.unparse(writer, -1, -1);
```
AndrewPochapsky, line 953: Why is this commented out? Is this case no longer necessary?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1332,9 +1337,10 @@ SqlSelect SqlSelect() :
         }
     )
     {
-        return new SqlSelect(s.end(this), keywordList,
+        return new SqlSelect(s.end(this), keywordList, topN,
             new SqlNodeList(selectList, Span.of(selectList).pos()),
-            fromClause, where, groupBy, having, qualify, windowDecls, null, null, null,
+            fromClause, where, groupBy, having, qualify, windowDecls,
+            /*orderBy=*/null, /*offset=*/null, /*fetch=*/null,
```
AndrewPochapsky, line 1343: nit: Add spaces between \*/ and null

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
dstekol: My bad. Yes, surprisingly enough the existing code relies on this method doing nothing.

babel/src/main/java/org/apache/calcite/sql/babel/SqlSelectTopN.java
dstekol: Will do.

babel/src/main/java/org/apache/calcite/sql/babel/SqlSelectTopN.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.SqlSpecialOperator;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+
+/**
+ * <code>SqlSelectTopN</code> is a class that handles the TOP n syntax in
```
ArthurInTheShell, line 34: nit: use {@code} wrap for the class name

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1268,6 +1268,7 @@ SqlSelect SqlSelect() :
 {
     final List<SqlLiteral> keywords = new ArrayList<SqlLiteral>();
     final SqlNodeList keywordList;
+    SqlNode topN = null;
```
ArthurInTheShell, line 1271: nit: wrap in <#if></#if>

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -111,7 +131,8 @@ public SqlSelect(SqlParserPos pos,
 
   @Override public List<SqlNode> getOperandList() {
     return ImmutableNullableList.of(keywordList, selectList, from, where,
-        groupBy, having, qualify, windowDecls, orderBy, offset, fetch, hints);
+        groupBy, having, qualify, windowDecls, orderBy, offset, fetch,
+        hints, topN);
```
ArthurInTheShell, line 135: It seems to me in the design doc, the discussion of the order of the operand does not align with the change you have here. You would either need to update the design doc, or move topN to the correct order in the code.

dstekol: I will update the design doc. It seems the order is not crucial to unparsing, so I am going with Drew's suggestion of minimizing the chances of breaking anything by adding to the end.

babel/src/main/java/org/apache/calcite/sql/babel/SqlSelectTopN.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(selectNum, isPercent, withTies);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("TOP");
+    selectNum.unparse(writer, leftPrec, rightPrec);
+    if (isPercent.getValueAs(Boolean.class)) {
```
dasch-1, line 63: Should we also handle the case where isPercent is null?

babel/src/main/java/org/apache/calcite/sql/babel/SqlSelectTopN.java
```diff
@@ -0,0 +1,70 @@
...
+      writer.keyword("PERCENT");
+    }
+    if (withTies.getValueAs(Boolean.class)) {
```
dasch-1, line 66: Should we also handle the case where withTies is null?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1307,6 +1308,10 @@ SqlSelect SqlSelect() :
     |   <ALL> {
             keywords.add(SqlSelectKeyword.ALL.symbol(getPos()));
         }
+<#if parser.includeTopN>
+    |
+        topN = SqlSelectTopN(getPos())
```
dasch-1, line 1313: So does this mean that the TOP N clause is mutually exclusive with the DISTINCT keyword?

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -949,7 +949,15 @@ public void unparseOffsetFetch(SqlWriter writer, SqlNode offset,
...
    */
-  public void unparseTopN(SqlWriter writer, SqlNode offset, SqlNode fetch) {
+  public void unparseFetchAsTopN(SqlWriter writer, SqlNode offset, SqlNode fetch) {
+  }
+
+  /**
+   * Unparses a TOP N clause (ex. SELECT TOP 5 * FROM FOO)
+   * @param topN
+   */
+  public void unparseTopN(SqlWriter writer, SqlNode topN) {
```
dasch-1, line 959: Should we check for a null topN here?

core/src/main/codegen/templates/Parser.jj
dstekol: Yes, this is what the syntax diagram says the behavior should be.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
dstekol: This check is done in SqlSelectOperator, but the unparseTopN could theoretically be called from somewhere else as well, so it probably does make sense to add it just to be safe.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1197,3 +1197,42 @@ SqlNode RankFunctionCallWithParams() :
         return SqlStdOperatorTable.OVER.createCall(s.end(over), rankCall, over);
     }
 }
+
```
lisayin, line 1200: Maybe add a top level comment?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1667,4 +1667,45 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test void testTopNDecimalWithoutPercentFails() {
+    final String sql = "select top ^5.2^ bar from foo";
+    final String expected = "(?s).*Cannot specify non-integer value."
+        + "*without specifying PERCENT.*";
+    sql(sql).fails(expected);
+  }
+
+  @Test void testTopNPercentGreaterThan100Fails() {
```
lisayin, line 1706: Add a test that sets N to be more than 100 but doesn't specify PERCENT

core/src/main/java/org/apache/calcite/sql/dialect/SybaseSqlDialect.java
```diff
@@ -36,11 +36,11 @@ public SybaseSqlDialect(Context context) {
 
   @Override public void unparseOffsetFetch(SqlWriter writer, SqlNode offset,
       SqlNode fetch) {
-    // No-op; see unparseTopN.
+    // No-op; see unparseFetchAsTopN.
```
lisayin, line 39: I might have missed it but where is FETCH NEXT n ROWS -> SELECT TOP(n) tested?

dstekol: I believe this is tested in the RelToSqlConverterTest file (for example in testSelectQueryWithFetchClause unit test). Converting FETCH statements into SELECT TOP N statements is not actually the functionality we are trying to add here, so the changes here are just to disentangle the TOP N unparsing logic from the FETCH unparsing logic (which sometimes ends up being unparsed as a TOP N statement as per pre-existing code).

lisayin: That makes sense

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1667,4 +1667,52 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "select top 200 ^percent^ bar from foo";
+    final String expected = "(?s).*Numeric literal.*out of range.*";
+    sql(sql).fails(expected);
+  }
```
michaelvdow, line 1717: Could add a test here for negative numbers to show that it should fail on that input

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1007,6 +1007,9 @@
   /** The {@code TUMBLE} group function. */
   TUMBLE,
 
+  /** The {@code TOP N} function*/
```
michaelvdow, line 1010: nit: period and space after "function"

core/src/main/java/org/apache/calcite/sql/SqlWriter.java
```diff
@@ -376,11 +376,16 @@ public String getName() {
...
-   * Prints the TOP(n) clause.
+   * Prints the fetch clause as a TOP(n) clause.
    *
    * @see #fetchOffset
    */
-  void topN(SqlNode fetch, SqlNode offset);
+  void fetchAsTopN(SqlNode fetch, SqlNode offset);
+
+  /**
+   * Prints the TOP N clause
```
michaelvdow, line 386: nit: period

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dstekol: Good call, will do.

PR #93, opened 2020-06-23
[CALCITE] Gradle task: Extraction of functions

AndrewPochapsky:
Tests will be added in a future PR as they will test the script as a whole, not each individual function.

### DIFF
--- a/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
+++ b/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
@@ -22,6 +22,7 @@ import java.util.Queue
 import javax.inject.Inject
 import kotlin.collections.HashMap
 import kotlin.collections.MutableMap
+import kotlin.text.MatchResult
 import kotlin.text.Regex
 import kotlin.text.StringBuilder
 import org.gradle.api.DefaultTask
@@ -46,12 +47,15 @@ open class DialectGenerateTask @Inject constructor(
     @TaskAction
     fun run() {
         val functionMap = extractFunctions()
+        functionMap.forEach {
+            k, v ->
+                println("$k = $v")
+            }
     }
 
     /**
      * Traverses the parsing directory structure and extracts all of the
-     * functions located in *.ftl files into a Map
-     *
+     * functions located in *.ftl files into a Map.
      */
     private fun extractFunctions(): Map<String, String> {
         val rootDirectoryFile = rootDirectory.get().asFile
@@ -104,8 +108,7 @@ open class DialectGenerateTask @Inject constructor(
         for (f in files) {
             if (f.isFile && f.extension == "ftl") {
                 processFile(f, functionMap)
-            }
-            if (directories.isNotEmpty() && f.name == nextDirectory) {
+            } else if (directories.isNotEmpty() && f.name == nextDirectory) {
                 directories.poll()
                 traverse(directories, f, functionMap)
             }
@@ -128,24 +131,66 @@ open class DialectGenerateTask @Inject constructor(
     private fun processFile(file: File, functionMap: MutableMap<String, String>) {
         var fileText = file.readText(Charsets.UTF_8)
         val typeAndName = "\\w+\\s+\\w+"
+        val delims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/))"
         val declarationPattern =
             Regex("(%s\\s*\\(\\s*(%s\\s*(\\,\\s*%s\\s*)*)?\\)\\s*\\:\n)"
                     .format(typeAndName, typeAndName, typeAndName))
-        val delims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/))"
         // Uses Lookahead and Lookbehind to ensure that the delims are added as tokens.
         val splitRegex = Regex("((?<=%s)|(?=%s))".format(delims, delims))
-        val declarationMatches = declarationPattern.findAll(fileText)
-        for (m in declarationMatches) {
+        val declarations: Queue<MatchResult> = LinkedList(declarationPattern.findAll(fileText).toList())
+        val tokens: Queue<String> = LinkedList(splitRegex.split(fileText))
+        var declaration = declarations.poll()
+        var charIndex = 0
+        if (declarations.isNotEmpty()) {
+            while (tokens.isNotEmpty()) {
+                if (charIndex == declaration.range.start) {
+                    charIndex = processFunction(tokens, functionMap, charIndex, declaration.range.endInclusive)
+                    if (declarations.isNotEmpty()) {
+                        declaration = declarations.poll()
+                    } else {
+                        return
+                    }
+                } else {
+                    val token = tokens.poll()
+                    charIndex += token.length
+                }
+            }
+        }
+        /*for (m in declarationMatches) {
             val functionDeclaration = m.value
             val functionName = getFunctionName(functionDeclaration)
             val functionBuilder = StringBuilder(functionDeclaration)
             val declarationIndex = fileText.indexOf(functionDeclaration)
             fileText = fileText.substring(declarationIndex + functionDeclaration.length)
-            val tokens: Queue<String> = LinkedList(splitRegex.split(fileText))
             processCurlyBlock(functionBuilder, tokens)
             processCurlyBlock(functionBuilder, tokens)
             functionMap.put(functionName, functionBuilder.toString())
+        }*/
+    }
+
+    private fun processFunction(
+        tokens: Queue<String>,
+        functionMap: MutableMap<String, String>,
+        charIndex: Int,
+        declarationEnd: Int
+    ): Int {
+        var updatedCharIndex = charIndex
+        val functionBuilder = StringBuilder()
+        // Process the declaration:
+        while (updatedCharIndex != declarationEnd && tokens.isNotEmpty()) {
+            val token = tokens.poll()
+            functionBuilder.append(token)
+            updatedCharIndex += token.length
         }
+
+        // Called twice as there are two curly blocks.
+        updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
+        updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
+
+        println("boi: $functionBuilder")
+        println()
+
+        return updatedCharIndex
     }
 
     /**
@@ -158,15 +203,17 @@ open class DialectGenerateTask @Inject constructor(
      * @param functionBuilder The builder unto which the tokens get added to once parsed
      * @param tokens The tokens to be parsed
      */
-    private fun processCurlyBlock(functionBuilder: StringBuilder, tokens: Queue<String>) {
+    private fun processCurlyBlock(functionBuilder: StringBuilder, tokens: Queue<String>, charIndex: Int): Int {
         var curlyCounter = 0
         var insideString = false
         var insideSingleLineComment = false
         var insideMultiLineComment = false
         var validBrace = true
+        var updatedCharIndex = charIndex
         while (tokens.isNotEmpty()) {
             val token = tokens.poll()
             functionBuilder.append(token)
+            updatedCharIndex += token.length
             validBrace = !insideString &&
                     !insideSingleLineComment &&
                     !insideMultiLineComment
@@ -191,9 +238,10 @@ open class DialectGenerateTask @Inject constructor(
                 curlyCounter--
             }
             if (curlyCounter == 0) {
-                return
+                return updatedCharIndex
             }
         }
+        return updatedCharIndex
     }
 
     /**
--- a/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
+++ b/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
@@ -31,9 +31,27 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.TaskAction
 
+/* Gradle task that traverses the parsing directory and generates the parserImpls.ftl
+  file for the given dialect.
+ */
 open class DialectGenerateTask @Inject constructor(
     objectFactory: ObjectFactory
 ) : DefaultTask() {
+
+    private val TYPE_AND_NAME = "\\w+\\s+\\w+"
+    private val SPLIT_DELIMS = "(\\s|\n|\"|(//)|(/\\*)|(\\*/)|')"
+
+    private val tokenizer = Regex("((?<=$SPLIT_DELIMS)|(?=$SPLIT_DELIMS))")
+    private val declarationPattern =
+        Regex("($TYPE_AND_NAME\\s*\\(\\s*($TYPE_AND_NAME\\s*(\\,\\s*$TYPE_AND_NAME\\s*)*)?\\)\\s*\\:(\n)?)")
+    private val nameRegex = Regex("\\w+")
+
+    // Flags used when parsing.
+    private var validBrace = false
+    private var validDoubleQuote = false
+    private var validSingleQuote = false
+    private var validSingleComment = false
+    private var validMultiComment = false
 
     @InputDirectory
     val dialectDirectory = objectFactory.directoryProperty()
@@ -109,6 +127,8 @@ open class DialectGenerateTask @Inject constructor(
             if (f.isFile && f.extension == "ftl") {
                 processFile(f, functionMap)
             } else if (directories.isNotEmpty() && f.name == nextDirectory) {
+                // Remove the front element in the queue, the value is used above with
+                // directories.peek().
                 directories.poll()
                 traverse(directories, f, functionMap)
             }
@@ -130,66 +150,73 @@ open class DialectGenerateTask @Inject constructor(
      */
     private fun processFile(file: File, functionMap: MutableMap<String, String>) {
         var fileText = file.readText(Charsets.UTF_8)
-        val typeAndName = "\\w+\\s+\\w+"
-        val delims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/))"
-        val declarationPattern =
-            Regex("(%s\\s*\\(\\s*(%s\\s*(\\,\\s*%s\\s*)*)?\\)\\s*\\:\n)"
-                    .format(typeAndName, typeAndName, typeAndName))
-        // Uses Lookahead and Lookbehind to ensure that the delims are added as tokens.
-        val splitRegex = Regex("((?<=%s)|(?=%s))".format(delims, delims))
         val declarations: Queue<MatchResult> = LinkedList(declarationPattern.findAll(fileText).toList())
-        val tokens: Queue<String> = LinkedList(splitRegex.split(fileText))
+        // In this case there aren't any functions in the file.
+        if (declarations.isEmpty()) {
+            return
+        }
+        val tokens: Queue<String> = LinkedList(tokenizer.split(fileText))
         var declaration = declarations.poll()
         var charIndex = 0
-        if (declarations.isNotEmpty()) {
-            while (tokens.isNotEmpty()) {
-                if (charIndex == declaration.range.start) {
-                    charIndex = processFunction(tokens, functionMap, charIndex, declaration.range.endInclusive)
-                    if (declarations.isNotEmpty()) {
-                        declaration = declarations.poll()
-                    } else {
-                        return
-                    }
+        while (tokens.isNotEmpty()) {
+            // Found a function declaration.
+            if (charIndex == declaration.range.start) {
+                charIndex = processFunction(tokens, functionMap, charIndex,
+                    declaration.range.endInclusive, getFunctionName(declaration.value))
+                if (declarations.isNotEmpty()) {
+                    declaration = declarations.poll()
                 } else {
-                    val token = tokens.poll()
-                    charIndex += token.length
+                    // Have parsed all functions.
+                    return
                 }
+            } else {
+                // Consume the token and update the character index.
+                val token = tokens.poll()
+                charIndex += token.length
             }
         }
-        /*for (m in declarationMatches) {
-            val functionDeclaration = m.value
-            val functionName = getFunctionName(functionDeclaration)
-            val functionBuilder = StringBuilder(functionDeclaration)
-            val declarationIndex = fileText.indexOf(functionDeclaration)
-            fileText = fileText.substring(declarationIndex + functionDeclaration.length)
-            processCurlyBlock(functionBuilder, tokens)
-            processCurlyBlock(functionBuilder, tokens)
-            functionMap.put(functionName, functionBuilder.toString())
-        }*/
     }
 
+    /**
+     * Parses a function of the form:
+     *
+     * <return_type> <name>(<args>) :
+     * {
+     *     <properties>
+     * }
+     * {
+     *     <body>
+     * }
+     *
+     * @param tokens The tokens starting from the function declaration and
+     *               ending at EOF
+     * @param functionMap The map to which the extracted functions will be added to
+     * @param charIndex The character index of the entire text of the file at which
+     *                  the parsing is commencing at
+     * @param declarationEnd The char index (of entire file text) at which the l
+     *                       function declaration ends
+     * @param functionName The name of the function being processed
+     */
     private fun processFunction(
         tokens: Queue<String>,
         functionMap: MutableMap<String, String>,
         charIndex: Int,
-        declarationEnd: Int
+        declarationEnd: Int,
+        functionName: String
     ): Int {
         var updatedCharIndex = charIndex
         val functionBuilder = StringBuilder()
         // Process the declaration:
-        while (updatedCharIndex != declarationEnd && tokens.isNotEmpty()) {
+        while (updatedCharIndex < declarationEnd && tokens.isNotEmpty()) {
             val token = tokens.poll()
             functionBuilder.append(token)
             updatedCharIndex += token.length
         }
-
         // Called twice as there are two curly blocks.
         updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
         updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
 
-        println("boi: $functionBuilder")
-        println()
-
+        functionMap.put(functionName, functionBuilder.toString())
         return updatedCharIndex
     }
 
@@ -202,36 +229,46 @@ open class DialectGenerateTask @Inject constructor(
      *
      * @param functionBuilder The builder unto which the tokens get added to once parsed
      * @param tokens The tokens to be parsed
+     * @param charIndex The character index of the entire text of the file at which
+     *                  the parsing is commencing at
      */
     private fun processCurlyBlock(functionBuilder: StringBuilder, tokens: Queue<String>, charIndex: Int): Int {
         var curlyCounter = 0
         var insideString = false
-        var insideSingleLineComment = false
-        var insideMultiLineComment = false
-        var validBrace = true
+        var insideCharacter = false
+        var insideSingleComment = false
+        var insideMultiComment = false
         var updatedCharIndex = charIndex
         while (tokens.isNotEmpty()) {
             val token = tokens.poll()
             functionBuilder.append(token)
             updatedCharIndex += token.length
-            validBrace = !insideString &&
-                    !insideSingleLineComment &&
-                    !insideMultiLineComment
-            if (token == "\n") {
-                if (insideSingleLineComment) {
-                    insideSingleLineComment = false
+            if (token == "\n" || token == " ") {
+                if (token == "\n") {
+                    if (insideSingleComment) {
+                        insideSingleComment = false
+                    }
+                    continue
                 }
-                // Allows for an abitrary amount of new lines to be parsed.
+                // Since new lines and spaces between curly blocks are valid in calcite,
+                // we want to allow an arbitrary number of them before we start
+                // keeping track of curly braces.
                 continue
             }
-            if (token == "\"" && !insideSingleLineComment && !insideMultiLineComment) {
+            // The following checks ensure that curly braces inside of strings
+            // and comments do not affect curlyCounter.
+            determineTokenValidity(insideString, insideCharacter,
+                    insideSingleComment, insideMultiComment)
+            if (token == "\"" && validDoubleQuote) {
                 insideString = !insideString
-            } else if (token == "//" && !insideMultiLineComment) {
-                insideSingleLineComment = true
-            } else if (token == "/*" && !insideString && !insideSingleLineComment) {
-                insideMultiLineComment = true
-            } else if (token == "*/" && !insideString && !insideSingleLineComment) {
-                insideMultiLineComment = false
+            } else if (token == "'" && validSingleQuote) {
+                insideCharacter = !insideCharacter
+            } else if (token == "//" && validSingleComment) {
+                insideSingleComment = true
+            } else if (token == "/*" && validMultiComment) {
+                insideMultiComment = true
+            } else if (token == "*/" && validMultiComment) {
+                insideMultiComment = false
             } else if (token == "{" && validBrace) {
                 curlyCounter++
             } else if (token == "}" && validBrace) {
@@ -251,8 +288,44 @@ open class DialectGenerateTask @Inject constructor(
      *                            <return_type> <name> (<args>) :
      */
     private fun getFunctionName(functionDeclaration: String): String {
-        val nameRegex = Regex("\\w+")
         val matches = nameRegex.findAll(functionDeclaration)
         return matches.elementAt(1).value
+    }
+
+    /**
+     * Sets the various flags which are used by processCurlyBloc() when
+     * parsing special characters.
+     *
+     * @param insideString Whether currently within a string " "
+     * @param insideCharacter Whether currently within a character ' '
+     * @param insideSingleLineComment Whether curently within single comment //
+     * @param insideMultiLineComment Whether currently within multi comment /*
+     */
+    private fun determineTokenValidity(
+        insideString: Boolean,
+        insideCharacter: Boolean,
+        insideSingleLineComment: Boolean,
+        insideMultiLineComment: Boolean
+    ) {
+        validBrace = !insideString &&
+                !insideSingleLineComment &&
+                !insideMultiLineComment &&
+                !insideCharacter
+
+        validDoubleQuote = !insideCharacter &&
+                !insideSingleLineComment &&
+                !insideMultiLineComment
+
+        validSingleQuote = !insideString &&
+                !insideSingleLineComment &&
+                !insideMultiLineComment
+
+        validSingleComment = !insideString &&
+                !insideCharacter &&
+                !insideMultiLineComment
+
+        validMultiComment = !insideString &&
+                !insideCharacter &&
+                !insideSingleLineComment
     }
 }
--- a/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
+++ b/buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
@@ -38,12 +38,12 @@ open class DialectGenerateTask @Inject constructor(
     objectFactory: ObjectFactory
 ) : DefaultTask() {
 
-    private val TYPE_AND_NAME = "\\w+\\s+\\w+"
-    private val SPLIT_DELIMS = "(\\s|\n|\"|(//)|(/\\*)|(\\*/)|')"
+    private val typeAndName = "\\w+\\s+\\w+"
+    private val splitDelims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/)|')"
 
-    private val tokenizer = Regex("((?<=$SPLIT_DELIMS)|(?=$SPLIT_DELIMS))")
+    private val tokenizer = Regex("((?<=$splitDelims)|(?=$splitDelims))")
     private val declarationPattern =
-        Regex("($TYPE_AND_NAME\\s*\\(\\s*($TYPE_AND_NAME\\s*(\\,\\s*$TYPE_AND_NAME\\s*)*)?\\)\\s*\\:(\n)?)")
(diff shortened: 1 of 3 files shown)


buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+        val declarationMatches = declarationPattern.findAll(fileText)
+        for (m in declarationMatches) {
+            val functionDeclaration = m.value
+            val functionName = getFunctionName(functionDeclaration)
+            val functionBuilder = StringBuilder(functionDeclaration)
+            val declarationIndex = fileText.indexOf(functionDeclaration)
+            fileText = fileText.substring(declarationIndex + functionDeclaration.length)
+            val tokens: Queue<String> = LinkedList(splitRegex.split(fileText))
+            processCurlyBlock(functionBuilder, tokens)
+            processCurlyBlock(functionBuilder, tokens)
```
michaelvdow, line 146: Should this be called only once?

AndrewPochapsky: It's called twice since it's parsing two curly blocks: the property declarations and the function body.

michaelvdow: Ah I see, that makes sense

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+    var outputFile = ""
+
+    @TaskAction
+    fun run() {
+        val functionMap = extractFunctions()
+    }
+
+    /**
+     * Traverses the parsing directory structure and extracts all of the
+     * functions located in *.ftl files into a Map
```
dasch-1, line 53: nit: missing punctuation

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+        currentDirectory: File,
+        functionMap: MutableMap<String, String>
+    ) {
+        val files = currentDirectory.listFiles()
+        // Ensures that files are processed first.
+        files.sortBy { it.isDirectory }
+        val nextDirectory = directories.peek()
+        for (f in files) {
+            if (f.isFile && f.extension == "ftl") {
+                processFile(f, functionMap)
```
dasch-1, line 106: Should we have a break in the body of this if or, alternatively, should the following if be an else-if?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+            }
+            if (directories.isNotEmpty() && f.name == nextDirectory) {
+                directories.poll()
```
dasch-1, line 109: What does poll() do? Suggest adding a comment explaining.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+     * }
+     * {
+     *     <body>
+     * }
+     * @param file The file to process
+     * @param functionMap The map to which the parsing functions will be added to
+     */
+    private fun processFile(file: File, functionMap: MutableMap<String, String>) {
+        var fileText = file.readText(Charsets.UTF_8)
+        val typeAndName = "\\w+\\s+\\w+"
```
dasch-1, line 130: Creating regex objects can be pretty expensive. Rather than defining/initializing typeAndName, declarationPattern, delims, and splitRegex locally for every file, can we just create the equivalent of a global version of each? Alternatively, can we create them in a higher level function and pass them in here?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /**
+     * Gets the function name from the declaration.
+     *
+     * @param functionDeclaration The function declaration of form
+     *                            <return_type> <name> (<args>) :
+     */
+    private fun getFunctionName(functionDeclaration: String): String {
+        val nameRegex = Regex("\\w+")
```
dasch-1, line 206: Same for this regex. Can we instantiate it in one location instead of for every function?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+                    .format(typeAndName, typeAndName, typeAndName))
+        val delims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/))"
+        // Uses Lookahead and Lookbehind to ensure that the delims are added as tokens.
+        val splitRegex = Regex("((?<=%s)|(?=%s))".format(delims, delims))
+        val declarationMatches = declarationPattern.findAll(fileText)
+        for (m in declarationMatches) {
+            val functionDeclaration = m.value
+            val functionName = getFunctionName(functionDeclaration)
+            val functionBuilder = StringBuilder(functionDeclaration)
+            val declarationIndex = fileText.indexOf(functionDeclaration)
```
dasch-1, line 142: Performing a regex search and then following that up with another linear string search for every function seems expensive and could become problematic for large parsing files. In the worst case, I think this is O(N+N*M)=>O(N*M) where N is the number of characters in the file and M is the number of functions.

Are there any optimizations we can make here? For example, should we do away with the regex altogether and just iterate through the text ourselves? Then, we should be able to bring this down to just O(N).

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
michaelvdow, line 53: nit: Missing period and extra new line

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+            fileText = fileText.substring(declarationIndex + functionDeclaration.length)
```
dasch-1, line 143: This substring operation could be expensive to. It will also require O(N*M) where N is the number of characters in the file and M is the number of functions.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @param functionBuilder The builder unto which the tokens get added to once parsed
+     * @param tokens The tokens to be parsed
+     */
+    private fun processCurlyBlock(functionBuilder: StringBuilder, tokens: Queue<String>) {
+        var curlyCounter = 0
+        var insideString = false
+        var insideSingleLineComment = false
+        var insideMultiLineComment = false
+        var validBrace = true
```
dasch-1, line 166: Since this variable is overwritten each iteration through the loop, let's just declare it there (if that's allowed by the language).

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+                    !insideSingleLineComment &&
+                    !insideMultiLineComment
+            if (token == "\n") {
+                if (insideSingleLineComment) {
+                    insideSingleLineComment = false
+                }
+                // Allows for an abitrary amount of new lines to be parsed.
+                continue
+            }
+            if (token == "\"" && !insideSingleLineComment && !insideMultiLineComment) {
```
dasch-1, line 180: Can strings be marked by any other characters besides a double quote? Is a single quote valid?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: Suggest adding a comment since that may not be immediately clear to a future reader.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Sure, poll just removes the front element and returns it.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Yep, I can move them to be global.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: That should be possible. I think we can iterate over it ourselves but only start "parsing" when we found a function declaration. So we can keep a temp string variable and keep updating it until it matches the declaration regex.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Will do.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Good point, this can probably be done away with if we change how we find function declarations (ie what you brought up in the previous comment).

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: Oh interesting. Why are we not using the returned value?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: I think single quotes are used for characters, I can add a similar case for it.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Good point, will change.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+    private fun traverse(
+        directories: Queue<String>,
+        currentDirectory: File,
+        functionMap: MutableMap<String, String>
+    ) {
+        val files = currentDirectory.listFiles()
+        // Ensures that files are processed first.
+        files.sortBy { it.isDirectory }
+        val nextDirectory = directories.peek()
+        for (f in files) {
```
ArthurInTheShell, line 104: nit: More vertical spacing to separate code into logical block would improve readability.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+import kotlin.collections.MutableMap
+import kotlin.text.Regex
+import kotlin.text.StringBuilder
+import org.gradle.api.DefaultTask
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.TaskAction
+
+open class DialectGenerateTask @Inject constructor(
```
lisayin, line 33: Add a top level comment about the class

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: We get the returned value above with directories.peek(). If we use directories.poll() there instead the function will not check the files within the dialect directory. I admit that this is a bit confusing so if you'd like I can try finding a more clear solution.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: Just adding my 2cents: I don't always find that extra newlines adds to code clarity

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+     */
+    private fun processCurlyBlock(functionBuilder: StringBuilder, tokens: Queue<String>) {
+        var curlyCounter = 0
+        var insideString = false
+        var insideSingleLineComment = false
+        var insideMultiLineComment = false
+        var validBrace = true
+        while (tokens.isNotEmpty()) {
+            val token = tokens.poll()
+            functionBuilder.append(token)
```
lisayin, line 169: Shouldn't the token be appended after all the checks are cleared instead of before?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: This is fine as long as it works. But definitely worth writing that in a comment for a future reader's benefit :)

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: It is added before since there is no case in which we want to not add a token to the builder. The checks are only there to check if we want to treat a curly brace as an actual curly brace.

dasch-1: Great catch, Lisa! Yes, I think this would be undesirable if the first token is, say, a newline.

However, I think the main assumption here is that the first character will be an open curly brace. Although if we're making that assumption, it would be nice to codify it with some sort of an assert.

lisayin: Does this function only get passed legitimate curly blocks?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.TaskAction
+
+open class DialectGenerateTask @Inject constructor(
+    objectFactory: ObjectFactory
+) : DefaultTask() {
+
+    @InputDirectory
+    val dialectDirectory = objectFactory.directoryProperty()
```
ArthurInTheShell, line 38: Would this be a constant? If so, is it reasonable to add the "const" in declaration?

AndrewPochapsky: I'm not sure if kotlin has consts. The val keyword is pretty much const though as once it is assigned it can't be changed.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
ArthurInTheShell: Okay then, that would be up to Andrew's discretion.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,210 @@
...
+     *     <body>
+     * }
+     * @param file The file to process
+     * @param functionMap The map to which the parsing functions will be added to
+     */
+    private fun processFile(file: File, functionMap: MutableMap<String, String>) {
+        var fileText = file.readText(Charsets.UTF_8)
+        val typeAndName = "\\w+\\s+\\w+"
+        val declarationPattern =
+            Regex("(%s\\s*\\(\\s*(%s\\s*(\\,\\s*%s\\s*)*)?\\)\\s*\\:\n)"
```
ArthurInTheShell, line 132: nit: Would it be reasonable to extract regex strings to constants that have names and therefore are self documented like the example in [ParenthesisBalancer.kt](https://github.com/googleinterns/calcite/blob/6197b2e3cc40a018aeab4b16af3678991ab025fa/buildSrc/subprojects/buildext/src/main/kotlin/org/apache/calcite/buildtools/buildext/dsl/ParenthesisBalancer.kt)

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
ArthurInTheShell: I think it is still up to you if you want to make this a constant. It depends on if you would want future contributor to think twice if they want to change the value of dialectDirectory in the middle of the class.

AndrewPochapsky: It shouldn't be possible to change it though since it is a val and not a var.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: This opens an interesting situation: do we assume that all of the functions in parserImpls.ftl are valid? If not, I can add some checks to ensure that it follows the assumed pattern.

AndrewPochapsky: Currently the if block that checks for new line allows for an arbitrary amount of newlines before the first curly brace as I wasn't sure if that was a valid format or not.

lisayin: I'm a little confused what this function does it if puts every token into the function builder regardless of any of the checks. Is it supposed to leave out any information after the last closing brace?

Will this be a valid input to the function?

> {
>  info
> }
> {
>   second info
> }

AndrewPochapsky: Yes, it would stop after the first closing brace. Currently the function is being called twice inside the processFile function in order to process both curly blocks.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: I like that approach, will do!

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
lisayin: ah that makes sense

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -127,25 +146,74 @@ open class DialectGenerateTask @Inject constructor(
...
-            val functionDeclaration = m.value
-            val functionName = getFunctionName(functionDeclaration)
-            val functionBuilder = StringBuilder(functionDeclaration)
-            val declarationIndex = fileText.indexOf(functionDeclaration)
-            fileText = fileText.substring(declarationIndex + functionDeclaration.length)
-            val tokens: Queue<String> = LinkedList(splitRegex.split(fileText))
-            processCurlyBlock(functionBuilder, tokens)
-            processCurlyBlock(functionBuilder, tokens)
-            functionMap.put(functionName, functionBuilder.toString())
+        val declarations: Queue<MatchResult> = LinkedList(declarationPattern.findAll(fileText).toList())
```
dasch-1, line 149: I think we may be able to simplify the logic in this function a bit. How about this alternative representation? (Disclaimer: there may be bugs :)

val declarations: Queue<MatchResult> = LinkedList(declarationPattern.findAll(fileText).toList())
val tokens: Queue<String> = LinkedList(tokenizer.split(fileText))
var charIndex = 0
while (declarations.isNotEmpty()) {
  declaration = declarations.poll()
  while (charIndex < declaration.range.start) {
    charIndex += tokens.poll().length
  }
  charIndex = processFunction(tokens, functionMap, charIndex,
      declaration.range.endInclusive, getFunctionName(declaration.value))
}

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -30,10 +31,28 @@ import org.gradle.api.tasks.Input
...
 
+/* Gradle task that traverses the parsing directory and generates the parserImpls.ftl
+  file for the given dialect.
+ */
 open class DialectGenerateTask @Inject constructor(
     objectFactory: ObjectFactory
 ) : DefaultTask() {
 
+    private val typeAndName = "\\w+\\s+\\w+"
+    private val splitDelims = "(\\s|\n|\"|(//)|(/\\*)|(\\*/)|')"
```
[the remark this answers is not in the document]
AndrewPochapsky: It is supposed to represent the double slash for single line comments. I'll add a comment explaining this though.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Very nice! With a few syntax fixes it work :)

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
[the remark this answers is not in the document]
AndrewPochapsky: I just realized that I can remove the parentheses around each deliminator which seems to improve the readability a fair bit. Should I still add a comment?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+     * @param functionName The name of the function being processed
+     */
+    private fun processFunction(
+        tokens: Queue<String>,
+        functionMap: MutableMap<String, String>,
+        charIndex: Int,
+        declarationEnd: Int,
+        functionName: String
+    ): Int {
+        var updatedCharIndex = charIndex
```
dasch-1, line 189: nit: Let's just use "charIndex" and update it within this function since it should be passed in by value by the caller given that it's a primitive data type.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+        val functionBuilder = StringBuilder()
+        // Process the declaration:
+        while (updatedCharIndex < declarationEnd && tokens.isNotEmpty()) {
+            val token = tokens.poll()
+            functionBuilder.append(token)
+            updatedCharIndex += token.length
+        }
+        // Called twice as there are two curly blocks.
```
dasch-1, line 197: nit: Let's be explicit about what these two curly blocks are in the comment: one for initialization block and one for body block

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+    private val typeAndName = "\\w+\\s+\\w+"
+    private val splitDelims = "(\\s|\n|\"|//|/\\*|\\*/|')"
+
+    private val tokenizer = Regex("((?<=$splitDelims)|(?=$splitDelims))")
+    private val declarationPattern =
+        Regex("($typeAndName\\s*\\(\\s*($typeAndName\\s*(\\,\\s*$typeAndName\\s*)*)?\\)\\s*\\:\n?)")
+    private val nameRegex = Regex("\\w+")
+
+    // Flags used when parsing.
+    private var validBrace = false
```
dasch-1, line 50: Should these be declared at the class level? Couldn't this cause unwanted spillover between functions/files that we process?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+
+    /**
+     * Parses a block of text surrounded by curly braces. The function keeps track
+     * of the number of curly braces that have yet to be closed. Once this counter
+     * reaches 0 the block has been fully parsed. The function also ensures that
+     * curly braces within single-line comments, multi-line comments, and strings
+     * are not counted for this.
+     *
+     * @param functionBuilder The builder unto which the tokens get added to once parsed
+     * @param tokens The tokens to be parsed
```
dasch-1, line 213: We should expand on this @param javadoc by saying that this represents all the tokens in the file from this function definition onwards, not necessarily just the tokens that will be considered part of this function.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+        // Called twice as there are two curly blocks.
+        updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
+        updatedCharIndex = processCurlyBlock(functionBuilder, tokens, updatedCharIndex)
+
+        functionMap.put(functionName, functionBuilder.toString())
+        return updatedCharIndex
+    }
+
+    /**
+     * Parses a block of text surrounded by curly braces. The function keeps track
```
dasch-1, line 206: We should say that this function assumes the function definition/content are in a valid format.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+        var insideCharacter = false
+        var insideSingleComment = false
+        var insideMultiComment = false
+        var updatedCharIndex = charIndex
+        while (tokens.isNotEmpty()) {
+            val token = tokens.poll()
+            functionBuilder.append(token)
+            updatedCharIndex += token.length
+            if (token == "\n" || token == " ") {
+                if (token == "\n") {
```
dasch-1, line 233: Nested if's are typically discouraged where it's reasonably possible to replace them with something else. Can we just split the "\n" and " " logic into two separate clauses to avoid this?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,316 @@
...
+     * reaches 0 the block has been fully parsed. The function also ensures that
+     * curly braces within single-line comments, multi-line comments, and strings
+     * are not counted for this.
+     *
+     * @param functionBuilder The builder unto which the tokens get added to once parsed
+     * @param tokens The tokens to be parsed
+     * @param charIndex The character index of the entire text of the file at which
+     *                  the parsing is commencing at
+     */
+    private fun processCurlyBlock(
```
dasch-1, line 217: From a high-level, I think this function (and its helper function determineFunctionValidity) could potentially benefit from some rearchitecting as it's a little difficult to grok.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: I had to do this since in Kotlin the values are passed in as "val" so they can't be modified.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: They are class level so I could create a the function determineTokenValidity() that sets them (for readability). There shouldn't be any spill over since these booleans are reset each time the curly block function is called.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: That's how it was initially but I thought this was better haha, will change back.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: I'll try to come up with something.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: Ah ok, good to know.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
dasch-1: Makes sense. However, I think making them class level also harms readability because it sends a message to a reader that they have a larger scope/lifetime than they actually should.

Suggest finding another way to do this along with the potential rearchitecture of processCurlyBlock().

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,386 @@
...
+         * @param token The token to parse
+         * @return Whether or not the curly block has been fully parsed
+         */
+        fun parseToken(token: String): Boolean {
+            if (token == " ") {
+                return false
+            }
+            if (token == "\n") {
+                // Single line comments are ended by a new line.
+                if (insideState == InsideState.SINGLE_COMMENT) {
```
michaelvdow, line 318: Could this statement go inside getUpdatedState and called here for the current token? This could keep all those types of state modification statements together

AndrewPochapsky: I agree that it would be better to have that handled inside of getUpdatedState. I'm going to try to come up with an approach that parses the useless spaces/newlines before processCurlyBlock is called.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,407 @@
...
+     * {
+     *     <body>
+     * }
+     *
+     * @param tokens The tokens starting from the function declaration and
+     *               ending at EOF
+     * @param functionMap The map to which the extracted functions will be added to
+     * @param charIndex The character index of the entire text of the file at which
+     *                  the parsing is commencing at
+     * @param declarationEnd The char index (of entire file text) at which the l
```
michaelvdow, line 178: nit: remove "l" at end of line?

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,407 @@
...
+        return updatedCharIndex
+    }
+
+    /**
+     * Parses a block of text surrounded by curly braces. Creates an instance of
+     * CurlyParser() that is used keep track of whether or not the curly block
+     * has been fully parsed or not. The current character index of the file text
+     * is updated and returned once parsing is complete.
+     *
+     * The function is assumed to be in a valid format.
```
michaelvdow, line 212: Is it assumed to be in valid format if it checks and throws an exception when it's not valid? Could add a @throw tag describing the exception

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -278,39 +276,132 @@ open class DialectGenerateTask @Inject constructor(
...
+         *
+         * for any token1, token2
+         * if parseToken(token1) is called and parseToken(token2) is called right after,
+         * then token2 comes directly after token1 in the stream of tokens.
+         *
+         * @param token The token to parse
+         * @return Whether or not the curly block has been fully parsed
+         */
+        fun parseToken(token: String): Boolean {
+            insideState = getUpdatedState(token)
```
dasch-1, line 338: nit: Suggest updating the state directly in getUpdatedState() since it has access to the insideState member variable and is already operating on it there. Also, consider renaming to just "updateState()".

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -278,39 +276,132 @@ open class DialectGenerateTask @Inject constructor(
...
+            return doneParsingCurlyBlock
+        }
+
+        /**
+         * Determines the updated state based on the token and current state.
+         *
+         * @param token The token to parse
+         * @return The updated state
+         */
+        private fun getUpdatedState(token: String): InsideState {
```
dasch-1, line 349: I really like the design of this function. IMO, this is a much more intuitive way of representing the state diagram nature of this algorithm. Nice work.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
```diff
@@ -0,0 +1,407 @@
...
+            when (token) {
+                "\n" -> {
+                    if (insideState == InsideState.SINGLE_COMMENT) {
+                        return InsideState.NONE
+                    }
+                    return insideState
```
dasch-1, line 355: If you decide to update this function to just set the insideState variable directly, you can omit all of the return statements. The ones that are currently returning something other than just "insideState" would instead just overwrite insideState with the new value.

buildSrc/subprojects/parser/src/main/kotlin/org/apache/calcite/buildtools/parser/DialectGenerateTask.kt
AndrewPochapsky: Good catch, I forgot to remove that.

PR #94, opened 2020-06-23
[CALCITE] Add support for INDEX

dstekol:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -604,19 +605,33 @@ SqlCreate SqlCreateTable(Span s, SqlCreateSpecifier createSpecifier) :
...
            if (index instanceof SqlPrimaryIndex) {
                primaryIndex = (SqlPrimaryIndex) index;
            }
-       }
-    )*
+           else{
+               indices.add(index);
+           }
+        }
+        (
```
dstekol, line 618: Is there a better way of ensuring that index declarations are comma separated (without the first declaration needing to be preceded by a comma)?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1535,7 +1535,7 @@ private void checkParseInfixCast(String sqlType) {
   }
 
   @Test public void testPrimaryIndexMultipleOverwritten() {
-    final String sql = "create table foo primary index bar (lname) primary index baz (fname)";
+    final String sql = "create table foo primary index bar (lname), primary index baz (fname)";
```
dstekol, line 1538: This is technically an edge case (since primary indices are only meant to be declared once), but if they were to be declared multiple times, they would need to be comma separated to conform to the syntax diagram, so I updated this test as well.

babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateTable.java
```diff
@@ -54,7 +56,8 @@ public SqlCreateTable(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
       SetType setType, Volatility volatility, boolean ifNotExists,
       SqlIdentifier name, List<SqlCreateAttribute> tableAttributes,
       SqlNodeList columnList, SqlNode query, WithDataType withData,
-      SqlPrimaryIndex primaryIndex, OnCommitType onCommitType) {
+      SqlPrimaryIndex primaryIndex, List<SqlIndex> indices,
+      OnCommitType onCommitType) {
```
dstekol, line 60: Since there is no code that relies on using this constructor without the "indices" parameter, I have modified it rather than creating a separate constructor to avoid bloating the class with unused constructors.

dasch-1: While there may not be any callers of this within the Calcite code base, this is not usually a safe assumption to make in general for what is effectively a public API. To be safe, we should add a new constructor rather than modifying the existing.

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: One thing that could simplify the parsing logic is to just put all indices, regardless of if they're primary or secondary, into the "indices" list that is passed into the SqlCreateTable constructor. Within the SqlCreateTable java code, you could enforce any constraints that you need to.

Then, you could break this portion of the parsing logic out into a helper function and simplify into something like:

void ExtractIndex(List<SqlIndex> indices):
{
}
{
   index = SqlCreateTableIndex(s)
   {
     indices.add(index);
   }
}

SqlCreate SqlCreateTable(Span s, SqlCreateSpecifier createSpecifier) :
{
...
  List<SqlIndex> indices = new ArrayList<SqlIndex>();
...
}
{
...
ExtractIndex(indices)
( COMMA ExtractIndex(indices) )*
...
}

dstekol: As per our conversation, I have adapted the parsing logic to add all indices to the list and then extract the primary ones (to minimize duplicated code), but have refrained from doing this inside the SqlCreateTable class so as to avoid reworking existing constructors.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -606,26 +606,21 @@ SqlCreate SqlCreateTable(Span s, SqlCreateSpecifier createSpecifier) :
...
-                   primaryIndex = (SqlPrimaryIndex) index;
-               }
-               else{
-                   indices.add(index);
-               }
-           }
+           <COMMA> index = SqlCreateTableIndex(s) { indices.add(index); }
         )*
+        {
+            //filter out any primary indices from index list
```
dasch-1, line 614: nit: Uppercase "F" and space after slashes. Also punctuation at the end.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1776,6 +1776,56 @@ private void checkParseInfixCast(String sqlType) {
...
+    sql(sql).ok(expected);
+  }
+
+  @Test void testPrimaryIndexWithSecondaryIndex() {
+    final String sql = "create table foo (bar integer, qux integer) "
+          + "primary index (bar), index (qux)";
+    final String expected = "CREATE TABLE `FOO` (`BAR` INTEGER, `QUX` INTEGER) "
+          + "PRIMARY INDEX (`BAR`), INDEX (`QUX`)";
+    sql(sql).ok(expected);
+
```
dasch-1, line 1828: Missing close bracket.

dstekol: Yeah, I must have accidentally deleted it when I was trying to fix conflicts.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -611,7 +611,7 @@ SqlCreate SqlCreateTable(Span s, SqlCreateSpecifier createSpecifier) :
            <COMMA> index = SqlCreateTableIndex(s) { indices.add(index); }
         )*
         {
-            //filter out any primary indices from index list
+            //Filter out any primary indices from index list.
```
dasch-1, line 614: nit: Still need the space after slashes :)

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -724,8 +734,7 @@ void FieldNameTypeCommaListWithoutOptionalNull(
 }
 
 /**
-    Parses an index declaration. Currently only supports PRIMARY INDEX statements,
-    but can be extended to support non-primary indices.
+    Parses an index declaration (both PRIMARY and non-primary indices).
```
ArthurInTheShell, line 737: nit: [comment style](https://google.github.io/styleguide/javaguide.html#s4.8.6-comments)

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -604,19 +605,28 @@ SqlCreate SqlCreateTable(Span s, SqlCreateSpecifier createSpecifier) :
...
-    )*
+    [
+        index = SqlCreateTableIndex(s) { indices.add(index); }
+        (
+           <COMMA> index = SqlCreateTableIndex(s) { indices.add(index); }
+        )*
+        {
+            // Filter out any primary indices from index list.
+            int i = 0;
+            while (i < indices.size() ) {
```
ArthurInTheShell, line 616: nit: extra space before closed parenthesis

PR #99, opened 2020-06-24
[CALCITE] Reorganized module Quidem tests into dialect specific tests

michaelvdow:
Notes:
 - Added SqlConformanceEnum values for dialects
 - DefaultQuidemTest runs all ".iq" files copied over from core/server
 - BabelQuidemTest divided into various dialects
 - Updated ParserImpl reference in each QuidemTest file


core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
```diff
@@ -77,11 +77,33 @@
 
   /** Conformance value that instructs Calcite to use SQL semantics
    * consistent with Microsoft SQL Server version 2008. */
-  SQL_SERVER_2008;
+  SQL_SERVER_2008,
 
+  /** Conformance value that instructs Calcite to use SQL semantics
```
lisayin, line 82: You might be missing BIG_QUERY here?

michaelvdow: BIG_QUERY should be referenced on line 49 in this file. For the other comment, I can make a parent "DialectQuidemTest" to remove some of the copied code

michaelvdow:
Yeah I think they should be failing as of now since these tests reference ParserImpls that don't yet exist because those will be built by the Gradle Task and build scripts. It might be fine to merge it even if failing since this isn't the master branch?

PR #103, opened 2020-06-25
[CALCITE] Date format column attribute in create table

ArthurInTheShell:


babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.calcite.sql.babel;
+
+import org.apache.calcite.sql.SqlColumnAttribute;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * A {@code SqlColumnAttributeDateFormat} is an AST node that gives
+ * a Date column of a specific format
```
AndrewPochapsky, line 26: nit: Punctuation.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
```diff
@@ -0,0 +1,50 @@
...
+   * @param formatString  The format string string literal
+   */
+  public SqlColumnAttributeDateFormat(SqlParserPos pos, SqlNode formatString) {
+    super(pos);
+    this.formatString = formatString;
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    if (formatString != null) {
```
AndrewPochapsky, line 45: I'm a bit confused as to how formatString can ever be null. Isn't this object only created when \<FORMAT\> \<string\> is parsed?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -874,6 +874,26 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testCreateTableDateFormatDateOnlyColumnLevelAttributes() {
```
AndrewPochapsky, line 877: nit: Attributes -> Attribute

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -874,6 +874,26 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "create table foo (bar date)";
+    final String expected = "CREATE TABLE `FOO` (`BAR` DATE)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testCreateTableDateFormatWithFormatStringColumnLevelAttributes() {
```
AndrewPochapsky, line 883: nit: Attributes -> Attribute

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * A {@code SqlColumnAttributeDateFormat} is an AST node that gives
+ * a Date column of a specific format
+ */
+public class SqlColumnAttributeDateFormat extends SqlColumnAttribute {
+
+  private SqlNode formatString = null;
```
dasch-1: private final

dasch-1: Also, no need to initialize this to null since it's set in the constructor.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Creates a {@code SqlCharacterSetToCharacterSet}.
```
dasch-1, line 33: Should this be SqlColumnAttributeDateFormat?

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
dasch-1: The way the code is currently laid out, I agree it can never be null. But as the parsing code changes over time, there are no guarantees of that. And having guards like this in place are usually a good idea to handle against future changes.

dasch-1: nit: To avoid extra nesting, suggest returning early if formatString == null.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: Good catch, fixed in https://github.com/googleinterns/calcite/pull/103/commits/759efab9843a94cc2467c3e2d47333e5b1c7ef5a

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dasch-1: This name is a little confusing to me because it contains "DateFormat", but the test case has no date format in it at all. It seems like maybe you're including that in the name because it's part of a date format PR, but it probably doesn't make sense to have it in there in a general context.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dasch-1: Here also, I would suggest renaming this to something similar to:

testCreateTableWithDateFormatStringColumnLevelAttribute()

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -874,6 +874,26 @@ private void checkParseInfixCast(String sqlType) {
...
+    final String sql = "create table foo (bar date format 'YYYY-MM-DD')";
+    final String expected = "CREATE TABLE `FOO` (`BAR` DATE FORMAT 'YYYY-MM-DD')";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testCreateTableDateFormatWithOtherColumnLevelAttributes() {
```
dasch-1, line 889: Here also, I would suggest renaming this to something similar to:

testCreateTableWithDateFormatStringAndOtherColumnLevelAttributes()

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: You are correct. The null check was a leftover piece of code from my previous trail approach. It is not needed here. Removed in https://github.com/googleinterns/calcite/pull/103/commits/759efab9843a94cc2467c3e2d47333e5b1c7ef5a.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Thanks, fixed in https://github.com/googleinterns/calcite/pull/103/commits/759efab9843a94cc2467c3e2d47333e5b1c7ef5a

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Thanks, fixed in https://github.com/googleinterns/calcite/pull/103/commits/759efab9843a94cc2467c3e2d47333e5b1c7ef5a.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
```diff
@@ -42,9 +41,7 @@ public SqlColumnAttributeDateFormat(SqlParserPos pos, SqlNode formatString) {
 
   @Override public void unparse(final SqlWriter writer, final int leftPrec,
       final int rightPrec) {
-    if (formatString != null) {
-      writer.keyword("FORMAT");
-      formatString.unparse(writer, 0, 0);
-    }
+    writer.keyword("FORMAT");
```
dasch-1, line 44: Suggest adding the null check back, but reversing the order such that you return early if it is null.

ArthurInTheShell: Actually, now when I look at the code. It seems like formatString could not be null, since we assigned the value from StringLiteraL() to it in ParserImpls.ftl

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/103/commits/90fafc4decfc6b71ddaa415ff891a422d4bc4fa0

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: Yes. Updated in https://github.com/googleinterns/calcite/pull/103/commits/90fafc4decfc6b71ddaa415ff891a422d4bc4fa0

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: Okay, will add guard in the next commit.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/103/commits/90fafc4decfc6b71ddaa415ff891a422d4bc4fa0

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
AndrewPochapsky: Can't another approach be to to this.formatString = Object.requireNotNull(formatString)? I'm not 100% that is the correct call but something like that.

ArthurInTheShell: Guard added back in https://github.com/googleinterns/calcite/pull/103/commits/b63554f4f67218e8e176cea3d14790c698612c18

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/103/commits/b63554f4f67218e8e176cea3d14790c698612c18

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeDateFormat.java
dasch-1: Yes, Objects.requireNonNull(...) would be a valid alternative (and possibly better).

PR #105, opened 2020-06-25
[CALCITE] Reorganized module JUnit tests into dialect specific tests

michaelvdow:
Notes:
 - Added build.gradle.kts file for each new dialect module
 - Core/server parser tests placed in parsing/SqlParserTest.java
 - babel parser test divided into dialects
 - settings.gradle updated to build new dialect modules
 - Some import and package names updated/fixed


AndrewPochapsky:
> For the paths that way org.apache.calcite.test, should this instead be org/apache/calcite/test?

michaelvdow:
Yeah, good catch! I thought that was the right name because IntelliJ combines these paths with dots in the folder view but they should all be separated.

PR #107, opened 2020-06-26
[CALCITE] Copied SQL Java classes from server/babel into core

michaelvdow:
Notes:
 - SqlCreateTable from babel is renamed to SqlCreateTableDialect1 since this class also exists in server


dasch-1:
This is too many files for my lazy self to review. So I'm going to just trust you on this :) Of course, we can always make fixes/adjustments in future PRs.

PR #108, opened 2020-06-26
[CALCITE] Exec macro param list

ArthurInTheShell:
Add support to the optional parameter list in Exec Macro statements.

### DIFF
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -788,13 +788,30 @@ SqlIndex SqlCreateTableIndex(Span s) :
 SqlNode SqlExecMacro() :
 {
     SqlIdentifier macro;
+    SqlNodeList paramVals = new SqlNodeList(getPos());
     Span s;
 }
 {
     macro = CompoundIdentifier() { s = span(); }
+    [SqlExecMacroArgument(paramVals)]
     {
-        return new SqlExecMacro(s.end(this), macro);
+        return new SqlExecMacro(s.end(this), macro, paramVals);
     }
+}
+
+void SqlExecMacroArgument(SqlNodeList paramVals) :
+{
+    SqlNode e;
+}
+{
+    <LPAREN>
+    (
+        e = Literal()
+        {
+            paramVals.add(e);
+        }
+    )+
+    <RPAREN>
 }
 
 SqlNode SqlUsingRequestModifier(Span s) :
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -794,7 +794,9 @@ SqlNode SqlExecMacro() :
 }
 {
     macro = CompoundIdentifier() { s = span(); }
-    [SqlExecMacroArgument(paramNames, paramValues)]
+    [
+        SqlExecMacroArgument(paramNames, paramValues)
+    ]
     {
         return new SqlExecMacro(s.end(this), macro, paramNames, paramValues);
     }
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -813,7 +813,16 @@ void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
         {
             paramValues.add(value);
         }
-    )+
+
+    )
+    (
+        <COMMA>
+        value = Literal()
+        {
+            paramValues.add(value);
+        }
+
+    )*
     <RPAREN>
     |
     <LPAREN>
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -837,7 +837,7 @@ void SqlExecMacroArgument(SqlNodeList params) :
 
 SqlNode SqlSimpleIdentifierEqualLiteral() :
 {
-    SqlNode name;
+    SqlIdentifier name;
     SqlNode value;
 }
 {
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -803,41 +803,50 @@ SqlNode SqlExecMacro() :
 
 void SqlExecMacroArgument(SqlNodeList params) :
 {
-    SqlNode name;
-    SqlNode value;
+    SqlNode e;
 }
 {
     <LPAREN>
     (
-        name = SimpleIdentifier()
-        <EQ>
-        value = Literal()
+        e = SqlSimpleIdentifierEqualLiteral()
         {
-            params.add(new SqlExecMacroParam(getPos(), name, value));
+            params.add(e);
         }
         (
             <COMMA>
-            name = SimpleIdentifier()
-            <EQ>
-            value = Literal()
+            e = SqlSimpleIdentifierEqualLiteral()
             {
-                params.add(new SqlExecMacroParam(getPos(), name, value));
+                params.add(e);
             }
         )*
     |
-        value = Literal()
+        e = Literal()
         {
-            params.add(new SqlExecMacroParam(getPos(), value));
+            params.add(new SqlExecMacroParam(getPos(), e));
         }
         (
             <COMMA>
-            value = Literal()
+            e = Literal()
             {
-                params.add(new SqlExecMacroParam(getPos(), value));
+                params.add(new SqlExecMacroParam(getPos(), e));
             }
         )*
     )
     <RPAREN>
+}
+
+SqlNode SqlSimpleIdentifierEqualLiteral() :
+{
+    SqlNode name;
+    SqlNode value;
+}
+{
+    name = SimpleIdentifier()
+    <EQ>
+    value = Literal()
+    {
+        return new SqlExecMacroParam(getPos(), name, value);
+    }
 }
 
 SqlNode SqlUsingRequestModifier(Span s) :
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -788,27 +788,36 @@ SqlIndex SqlCreateTableIndex(Span s) :
 SqlNode SqlExecMacro() :
 {
     SqlIdentifier macro;
-    SqlNodeList paramVals = new SqlNodeList(getPos());
+    SqlNodeList paramNames = new SqlNodeList(getPos());
+    SqlNodeList paramValues = new SqlNodeList(getPos());
     Span s;
 }
 {
     macro = CompoundIdentifier() { s = span(); }
-    [SqlExecMacroArgument(paramVals)]
+    [SqlExecMacroArgument(paramNames, paramValues)]
     {
-        return new SqlExecMacro(s.end(this), macro, paramVals);
+        return new SqlExecMacro(s.end(this), macro, paramNames, paramValues);
     }
 }
 
-void SqlExecMacroArgument(SqlNodeList paramVals) :
+void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
 {
-    SqlNode e;
+    SqlNode name;
+    SqlNode value;
 }
 {
     <LPAREN>
     (
-        e = Literal()
+        [
+        name = SimpleIdentifier()
+        <EQ>
+            {
+                paramNames.add(name);
+            }
+        ]
+        value = Literal()
         {
-            paramVals.add(e);
+            paramValues.add(value);
         }
     )+
     <RPAREN>
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -817,7 +817,7 @@ void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
     )
     (
         <COMMA>
-        value = Literal()
+        value = Expression(ExprContext.ACCEPT_SUB_QUERY)
         {
             paramValues.add(value);
         }
@@ -832,11 +832,23 @@ void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
         {
             paramNames.add(name);
         }
-        value = Literal()
+        value = Expression(ExprContext.ACCEPT_SUB_QUERY)
         {
             paramValues.add(value);
         }
-    )+
+    )
+    (
+        <COMMA>
+        name = SimpleIdentifier()
+        <EQ>
+        {
+            paramNames.add(name);
+        }
+        value = Expression(ExprContext.ACCEPT_SUB_QUERY)
+        {
+            paramValues.add(value);
+        }
+    )*
     <RPAREN>
 }
 
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -808,7 +808,6 @@ void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
     SqlNode value;
 }
 {
-        LOOKAHEAD(2)
         <LPAREN>
         (
             value = Literal()
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -836,7 +836,7 @@ void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
     |
         <LPAREN>
         (
-            value = Literal()
+            value = Expression(ExprContext.ACCEPT_SUB_QUERY)
             {
                 paramValues.add(value);
             }
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -807,15 +807,14 @@ void SqlExecMacroArgument(SqlNodeList params) :
     SqlNode value;
 }
 {
-        <LPAREN>
-        (
-            name = SimpleIdentifier()
-            <EQ>
-            value = Literal()
-            {
-                params.add(new SqlExecMacroParam(getPos(), name, value));
-            }
-        )
+    <LPAREN>
+    (
+        name = SimpleIdentifier()
+        <EQ>
+        value = Literal()
+        {
+            params.add(new SqlExecMacroParam(getPos(), name, value));
+        }
         (
             <COMMA>
             name = SimpleIdentifier()
@@ -825,25 +824,20 @@ void SqlExecMacroArgument(SqlNodeList params) :
                 params.add(new SqlExecMacroParam(getPos(), name, value));
             }
         )*
-        <RPAREN>
     |
-        <LPAREN>
-        (
-            value = Literal()
-            {
-                params.add(new SqlExecMacroParam(getPos(), value));
-            }
-
-        )
+        value = Literal()
+        {
+            params.add(new SqlExecMacroParam(getPos(), value));
+        }
         (
             <COMMA>
             value = Literal()
             {
                 params.add(new SqlExecMacroParam(getPos(), value));
             }
-
         )*
-        <RPAREN>
+    )
+    <RPAREN>
 }
 
 SqlNode SqlUsingRequestModifier(Span s) :
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -788,21 +788,20 @@ SqlIndex SqlCreateTableIndex(Span s) :
 SqlNode SqlExecMacro() :
 {
     SqlIdentifier macro;
-    SqlNodeList paramNames = new SqlNodeList(getPos());
-    SqlNodeList paramValues = new SqlNodeList(getPos());
+    SqlNodeList params = new SqlNodeList(getPos());
     Span s;
 }
 {
     macro = CompoundIdentifier() { s = span(); }
     [
-        SqlExecMacroArgument(paramNames, paramValues)
+        SqlExecMacroArgument(params)
     ]
     {
-        return new SqlExecMacro(s.end(this), macro, paramNames, paramValues);
+        return new SqlExecMacro(s.end(this), macro, params);
     }
 }
 
-void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
+void SqlExecMacroArgument(SqlNodeList params) :
 {
     SqlNode name;
     SqlNode value;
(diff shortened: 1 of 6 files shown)


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,70 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+    (
+        <COMMA>
+        value = Expression(ExprContext.ACCEPT_SUB_QUERY)
+        {
+            paramValues.add(value);
+        }
+
+    )*
+    <RPAREN>
+    |
```
michaelvdow, line 827: It could help readability to wrap everything for this "or" statement in parentheses and indent everything besides the "|" over one tab

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,70 @@ SqlIndex SqlCreateTableIndex(Span s) :
 SqlNode SqlExecMacro() :
 {
     SqlIdentifier macro;
+    SqlNodeList paramNames = new SqlNodeList(getPos());
+    SqlNodeList paramValues = new SqlNodeList(getPos());
     Span s;
 }
 {
     macro = CompoundIdentifier() { s = span(); }
+    [SqlExecMacroArgument(paramNames, paramValues)]
```
michaelvdow, line 797: nit: space between "[" brackets and statement

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,70 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
     }
 }
 
+void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
+{
+    SqlNode name;
+    SqlNode value;
+}
+{
+    LOOKAHEAD(2)
```
michaelvdow, line 809: I think the default for babel is a lookahead of 2. Does it work without this statement?

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Good point. Fixed in https://github.com/googleinterns/calcite/pull/108/commits/6bc3ba6bc336c390bc38e8c7508b7af19dbc7ee2

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Put in different line in https://github.com/googleinterns/calcite/pull/108/commits/7896e626fa4d16a2a8ac9b734f38f00f5cbfb6d2

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Thanks for the information. Removed LOOKAHEAD in https://github.com/googleinterns/calcite/pull/108/commits/1569d122e19bac9c710f4124eb031b942fc9ae12

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,71 @@ SqlIndex SqlCreateTableIndex(Span s) :
 SqlNode SqlExecMacro() :
 {
     SqlIdentifier macro;
+    SqlNodeList paramNames = new SqlNodeList(getPos());
```
dasch-1, line 791: Prefer using a class object to encapsulate both name and value.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,71 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+    [
+        SqlExecMacroArgument(paramNames, paramValues)
+    ]
     {
-        return new SqlExecMacro(s.end(this), macro);
+        return new SqlExecMacro(s.end(this), macro, paramNames, paramValues);
     }
 }
 
+void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
```
dasch-1, line 805: Does this function require that all params be of the same type (e.g. all with a named value or all without)? Is that a requirement according to the docs?

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -32,31 +33,63 @@
 import java.util.Objects;
 
 /**
- * Parse tree for {@code EXEC} statement.
+ * Parse tree for {@code SqlExecMacro} statement.
  */
 public class SqlExecMacro extends SqlCall implements SqlExecutableStatement {
   public static final SqlSpecialOperator OPERATOR =
       new SqlSpecialOperator("EXECUTE", SqlKind.OTHER);
```
dasch-1, line 40: While we're at it, let's create a distinct SqlKind for EXECUTE.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -32,31 +33,63 @@
...
   @Override public List<SqlNode> getOperandList() {
-    return ImmutableNullableList.of(name);
+    // the list of paramNames could be empty
+    return ImmutableNullableList.of(name, paramValues, paramNames);
   }
 
   @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
     writer.keyword("EXECUTE");
     name.unparse(writer, leftPrec, rightPrec);
+    if (paramValues.size() == 0) {
```
dasch-1, line 74: nit: We can replace this with the following to be a bit more explicit:

SqlNodeList.isEmptyList(paramValues);

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -32,31 +33,63 @@
...
+      return;
+    }
+    SqlWriter.Frame frame = writer.startList("(", ")");
+    if (paramNames.size() == 0) {
```
dasch-1, line 78: nit: Let's use SqlNodeList.isEmptyList() here too.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
```diff
@@ -32,31 +33,63 @@
...
+    } else {
+      for (int i = 0; i < paramNames.size(); i++) {
+        writer.sep(",", false);
+        paramNames.get(i).unparse(writer, 0, 0);
+        writer.keyword("=");
+        paramValues.get(i).unparse(writer, 0, 0);
+      }
+    }
+    writer.endList(frame);
+
```
dasch-1, line 92: nit: remove extra newline

babel/src/main/codegen/includes/parserImpls.ftl
dstekol: On a related note, it would be good to add a test case where some of the params are named and others arent, for example: execute foo (1, secondparam = 'Hello')
Whether this should fail or pass depends on specs.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,71 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+            name = SimpleIdentifier()
+            <EQ>
+            {
+                paramNames.add(name);
+            }
+            value = Expression(ExprContext.ACCEPT_SUB_QUERY)
+            {
+                paramValues.add(value);
+            }
+        )
```
dstekol, line 840: No need for parentheses (lines 830 & 840)

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,71 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+            <EQ>
+            {
+                paramNames.add(name);
+            }
+            value = Expression(ExprContext.ACCEPT_SUB_QUERY)
+            {
+                paramValues.add(value);
+            }
+        )*
+        <RPAREN>
```
dstekol, line 853: Since you have repeated logic, it may be worth considering extracting it to a separate function, passing the paramNames and paramValues list to that function, and have it handle adding things to the lists. Then you could have 
something like this (not necessarily with same func name):
extractParamPair(paramNames, paramValues)
( 
  <COMMA> 
  extractParamPair(paramNames, paramValues)
)*
This could also be useful later down the line if it turns out this kind of "paramName = paramValue" syntax appears in other contexts.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: This function does require all the params to be either all with a named value or all without. I would add more explanation in the design doc.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Since this is more of a design decision, let's discuss it in the design doc.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
ArthurInTheShell: I have added EXECUTE to the set of DML operator list. Base on the name EXEC MACRO, I believe it would fall in to the DML category. https://github.com/googleinterns/calcite/pull/108/commits/52b2158d1128394efb3ef1a7f723e3f23fecc648

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/108/commits/9dfb17bc583a57607a20110e68aeeb5541e30da5

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/108/commits/9dfb17bc583a57607a20110e68aeeb5541e30da5

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacro.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/108/commits/20c4b40c220d3429b0f091cd09306a0e0043d473

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,15 +788,71 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+}
+{
+        <LPAREN>
+        (
+            value = Literal()
+            {
+                paramValues.add(value);
+            }
+
+        )
```
dstekol, line 818: Also no need for these parentheses (812 & 818), and extra space on 817.

ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/108/commits/41eff94023a11dbee3d010b4db619eae40217431

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: This is a good point. Now the code does not work on two separated list anymore, I think it would be reasonable to leave it as it is. There are still few token that are repeated. However since only three token are repeated, contributors can extract the logic when a future feature needs it.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Removed parentheses in https://github.com/googleinterns/calcite/pull/108/commits/41eff94023a11dbee3d010b4db619eae40217431

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,69 +788,56 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+        SqlExecMacroArgument(params)
     ]
     {
-        return new SqlExecMacro(s.end(this), macro, paramNames, paramValues);
+        return new SqlExecMacro(s.end(this), macro, params);
     }
 }
 
-void SqlExecMacroArgument(SqlNodeList paramNames, SqlNodeList paramValues) :
+void SqlExecMacroArgument(SqlNodeList params) :
```
dasch-1, line 804: It seems like this function could be simplified by removing the "|" and having the following be optional in the top clause:

[
  name = SimpleIdentifier()
  "EQ"
]

And then a single constructor could be used, with name passed in optionally as null.

dasch-1: Furthermore, I agree with @dstekol  that this pattern could be extracted into a helper parsing function so as to remove as much duplication as possible from the initial case and the case within ()*.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,87 @@
...
+  private final SqlNode name;
+  private final SqlNode value;
+
+  /**
+   * Create an {@code SqlExecMacroParam} without name.
+   *
+   * @param pos  Parser position, must not be null
+   * @param value  Value of the parameter
+   */
+  public SqlExecMacroParam(SqlParserPos pos, SqlNode value) {
```
dasch-1, line 46: Can we get rid of this constructor and just use the one below with "name" optionally null?

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,87 @@
...
+    this.name = name;
+    this.value = value;
+  }
+
+  @Override public SqlOperator getOperator() {
+    return OPERATOR;
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    if (this.name == null) {
```
dasch-1, line 70: We should return the same size operand list in all cases, just with the zeroth element set to null possibly. Otherwise, a caller won't know if the zeroth argument is the name or the value.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,87 @@
...
+      return ImmutableNullableList.of(value);
+    } else {
+      return ImmutableNullableList.of(name, value);
+    }
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    if (this.name == null) {
```
dasch-1, line 79: Rather than replicate value.unparse(), in this case I think it makes more sense to wrap name.unparse() and writer.keyword("=") inside the if statement.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Regarding your first comment, if we simplified the pattern to the way you suggested, we will be allowing mix and match for parameter with and without name to be parsed in the same list. Which is not the desired behavior. My thought is not to delay the validation till runtime. Any suggestions?

ArthurInTheShell: Regarding your second comment, since two of you all pointed out extracting the pattern would be preferable. I am going to extract the pattern.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
ArthurInTheShell: Code wise, getting rid of this constructor would look like removing duplicate. However, the point of having two constructor taking different number of parameters is to enforce the name to be null internally. We don't need to rely on the future SWE to follow the standard of passing a null for name when initializing one without name. Please let me know what you think. If the pro is greater than the con, we should still get  rid of the constructor.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
ArthurInTheShell: That is a good point. Will fix soon.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/108/commits/b78afe5b0a41b4f4f2fc51d16f72d3f50d3b0269

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: That's a good point that blending the two would allow mix-and-match parameters, which we don't want.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
dasch-1: I think either way is fine, so I'll leave it up to you to decide whichever you'd prefer.

However, if you do choose to keep two different constructors, you should only have one main constructor which the others call. So in this case, that would mean having the SqlExecMacroParam(pos, value) constructor call the SqlExecMacroParam(pos, name, value) one.

ArthurInTheShell: Good suggestion. Will do.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Pattern extracted in https://github.com/googleinterns/calcite/pull/108/commits/823f91878be465524acca95e35c28db65f20a970

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+
+/**
+ * Parse tree for {@code SqlExecMacroParam} SqlCall.
+ */
+public class SqlExecMacroParam extends SqlCall {
+  private static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("PARAM EQUAL", SqlKind.OTHER);
+
+  private final SqlNode name;
```
dasch-1, line 37: Let's make this the more specific SqlIdentifier type.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,79 @@
...
+  private final SqlNode value;
```
dasch-1, line 38: Let's make this the more specific SqlLiteral type.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,12 +788,64 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+                params.add(new SqlExecMacroParam(getPos(), e));
+            }
+        )*
+    )
+    <RPAREN>
+}
+
+SqlNode SqlSimpleIdentifierEqualLiteral() :
+{
+    SqlNode name;
```
dasch-1, line 840: Let's make this the more specific SqlIdentifier type.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -788,12 +788,64 @@ SqlIndex SqlCreateTableIndex(Span s) :
...
+    SqlNode value;
```
dasch-1, line 841: Let's make this the more specific SqlLiteral type.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
ArthurInTheShell: Change made in https://github.com/googleinterns/calcite/pull/108/commits/e667292f4b74ef3f262506850fcf3cf6b113988e

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Change made in https://github.com/googleinterns/calcite/pull/108/commits/e667292f4b74ef3f262506850fcf3cf6b113988e

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
ArthurInTheShell: Since [Literal() function](https://github.com/googleinterns/calcite/blob/5d283ceedf9563bff1c715e1393fa15d094255f2/core/src/main/codegen/templates/Parser.jj#L4207) in Parser.jj returns SqlNode, it would be reasonable to use SqlNode here. Unless we want to do type casting to make it more specific.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Since [Literal() function](https://github.com/googleinterns/calcite/blob/5d283ceedf9563bff1c715e1393fa15d094255f2/core/src/main/codegen/templates/Parser.jj#L4207) in Parser.jj returns SqlNode, it would be reasonable to use SqlNode here. Unless we want to do type casting to make it more specific.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
dasch-1: Looks like you're right. SqlNode seems like the wrong return value for the Literal() function. But since that's not introduced in this change, I think it's fine to leave as is.

babel/src/main/codegen/includes/parserImpls.ftl
dasch-1: Looks like you're right. SqlNode seems like the wrong return value for the Literal() function. But since that's not introduced in this change, I think it's fine to leave as is.

babel/src/main/java/org/apache/calcite/sql/babel/SqlExecMacroParam.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(name, value);
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    if (this.name != null) {
+      name.unparse(writer, 0, 0);
+      writer.keyword("=");
```
dstekol, line 76: I believe this should be unparsed as a separator:
writer.sep("=");

ArthurInTheShell: It's good that you brought this to my attention. I have done some investigation into whether "=" is a separator or a keyword. Judging by [the Java doc of sep(String)](https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/SqlWriter.java#L493), it seems like sep(String) are used for list separator. I think "=" would be more of a keyword than a separator in this standard. Would you agree with my judgement or is there something I missed?

dstekol: I have mostly seen "=" unparsed as a separator in existing code. That said, if it is ambiguous and you feel this is more appropriate, that's fine too.

ArthurInTheShell: I would appreciate it, if you can point me to a few examples on where past contributor use "=" as separator.

PR #109, opened 2020-06-26
[CALCITE] Converted gradle task logic to java. 

AndrewPochapsky:
Converted gradle task to java and added tests. Generation logic will be added in a subsequent PR.


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+
+  private final String typeAndName = "\\w+\\s+\\w+";
+  private final String splitDelims = "(\\s|\n|\"|//|/\\*|\\*/|'|\\}|\\{)";
+
+  private final Pattern tokenizerPattern = Pattern.compile("((?<="
+      + splitDelims + ")|(?=" + splitDelims + "))");
+  private final Pattern declarationPattern =
+    Pattern.compile("(" + typeAndName + "\\s*\\(\\s*(" + typeAndName + "\\s*(\\,\\s*"
+        + typeAndName + "\\s*)*)?\\)\\s*\\:\n?)");
+  private final Pattern namePattern = Pattern.compile("\\w+");
```
lisayin, line 48: Can you add a comment with examples of what these regexes should match to?

AndrewPochapsky:
> Is there a test file that should be included with this?

That will be coming the next PR which will contain tests and the generation logic.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+import java.util.Map;
+import java.util.Queue;
+import java.util.LinkedList;
+import java.util.LinkedHashMap;
+import java.util.Queue;
+import java.util.regex.Matcher;
+import java.util.regex.MatchResult;
+import java.util.regex.Pattern;
+
+public class DialectGenerate {
```
lisayin, line 38: Top level comment on what this class does?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+
+  /**
+   * Extracts functions and generates parserImpls.ftl for the given dialect.
+   */
+  public void run() {
+    Map<String, String> functions = extractFunctions();
+    for (Map.Entry<String, String> entry : functions.entrySet()) {
+      String key = entry.getKey();
+      String value = entry.getValue();
+      System.out.println(key + "=" + value + "\n");
```
lisayin, line 68: Comment says that it generates parserImpls.ftl, but code seems to suggest that all it does is print to console.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+    traverse(queue, rootDirectory, functionMap);
+    return functionMap;
+  }
+
+  /**
+   * Generates the parserImpls.ftl file for the dialect.
+   *
+   * @param functions The functions to place into the output file
+   */
+  public void generateParserImpls(Map<String, String> functions) {}
```
lisayin, line 88: If you're not implementing it in this pr, then maybe add a TODO?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+      @Override
+      public int compare(File file1, File file2) {
+        return Boolean.compare(file1.isDirectory(), file2.isDirectory());
+      }
+    });
+    String nextDirectory = directories.peek();
+    for (File f : files) {
+      String extension = "";
+      String fileName = f.getName();
+      int i = fileName.lastIndexOf('.');
```
lisayin, line 129: Maybe add a general comment about if the file doesn't have an extension? Alternatively, `continue` out of the current loop if "." is not found.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+   * @param currentDirectory The current directory the function is processing
+   * @param functionMap The map to which the parsing functions will be added to
+   */
+  private void traverse(
+      Queue<String> directories,
+      File currentDirectory,
+      Map<String, String> functionMap) {
+    File[] files = currentDirectory.listFiles();
+    // Ensures that files are processed first.
+    Arrays.sort(files, new Comparator<File>() {
```
lisayin, line 119: Define this comparator outside of this function so it doesn't get created every time this function is run.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Adding a continue wouldn't work as then directories wouldn't get traversed.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: It will be in the next PR.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Will do

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: I'll add a comment explaining this

dasch-1: Agree with @lisayin that we should have the PR make sense standalone as much as possible. Can we have the comment accurately reflect that we only print for now. And in the follow up PR change it to reflect the new behavior?

Alternatively, you could leave the comment as is and write a "TODO(apochapsky): <description_of_what_to_do>" comment to make it clear to a reader/reviewer that there is an intention to change the logic.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+        return Boolean.compare(file1.isDirectory(), file2.isDirectory());
+      }
+    });
+    String nextDirectory = directories.peek();
+    for (File f : files) {
+      String extension = "";
+      String fileName = f.getName();
+      int i = fileName.lastIndexOf('.');
+      if (i > 0) {
+        extension = fileName.substring(i+1);
```
dasch-1, line 144: nit: /i+1/i + 1/, to be consistent with style used in rest of the file (e.g. line 178).

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+      }
+      if (f.isFile() && extension.equals("ftl")) {
+        try {
+          processFile(f, functionMap);
+        } catch(IOException e) {
+          e.printStackTrace();
+        }
+      } else if (!directories.isEmpty() && fileName.equals(nextDirectory)) {
+        // Remove the front element in the queue, the value is used above with
```
dasch-1, line 153: nit: This comment should clarify that the value is used above in the next recursive call to this function.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+   * has been fully parsed or not. The current character index of the file text
+   * is updated and returned once parsing is complete.
+   *
+   * @param functionBuilder The builder unto which the tokens get added to once parsed
+   * @param tokens The tokens starting from the function declaration and
+   *               ending at EOF
+   * @param charIndex The character index of the entire text of the file at which
+   *                  the parsing is commencing at
+   */
+  private int processCurlyBlock(
```
dasch-1, line 249: Should we update this function signature to indicate that it throws an IllegalStateException?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+      String extension = "";
+      String fileName = f.getName();
+      int i = fileName.lastIndexOf('.');
+      if (i > 0) {
+        extension = fileName.substring(i+1);
+      }
+      if (f.isFile() && extension.equals("ftl")) {
+        try {
+          processFile(f, functionMap);
+        } catch(IOException e) {
```
dasch-1, line 149: Is IOException the only exception that could be thrown in this call? It looks like IllegalStateException is also an option.

dasch-1, line 149: nit: Space between "catch" and open paren.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,403 @@
...
+        if (insideState == InsideState.SINGLE_COMMENT) {
+          insideState = InsideState.NONE;
+        }
+      } else if (token.equals("\"")) {
+        if (insideState == InsideState.NONE) {
+          insideState = InsideState.STRING;
+        } else if (insideState == InsideState.STRING) {
+          insideState = InsideState.NONE;
+        }
+      } else if (token.equals("'")) {
```
lisayin, line 387: what about ' tokens in strings? eg " something 'single quotes' something "

lisayin: actually nvm I think the previous case should cover it.

lisayin: Can you add a set of tests in the PR for just this CurlyParser? It's pretty standalone.

AndrewPochapsky: That is handled already by the state logic, I have a state diagram in my design doc which may help.

dasch-1: @AndrewPochapsky and I spoke offline. The plan is to add tests for the entire class in this PR, rather than a follow-up one.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,416 @@
...
+     * All of the possible important structures that a token can be inside of.
+     * This is tracked to ensure that when a curly brace is encountered,
+     * whether it is part of the syntax or not can be determined.
+     */
+    enum InsideState {
+
+      /**
+       * Not inside of anything.
+       */
+      NONE,
```
lisayin, line 325: This is kind of a misleading name as it can be either not inside anything or inside curly braces but not inside the other states.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,416 @@
...
+   * }
+   * {
+   *     <body>
+   * }
+   *
+   * @param file The file to process
+   * @param functionMap The map to which the parsing functions will be added to
+   */
+  private void processFile(File file, Map<String, String> functionMap) throws IOException {
+    String fileText = new String(Files.readAllBytes(Paths.get(file.getAbsolutePath())),
```
dasch-1, line 176: We should try to extract the Files.readAllBytes() out of this function and have this function instead just take the file contents as input.

That would make it a lot more straightforward to unit-test this function without having to do file I/O.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * Retrns a DialectGenerate with root path of calcite/parsingTest and
+   * dialect path of calcite/parsingTest/intermediate/testDialect
+   */
+  private DialectGenerate setupDialectGenerate() {
+    Path currentRelativePath = Paths.get("");
+    String s = currentRelativePath.toAbsolutePath().toString();
+    Path rootPath = Paths.get("..", "..", "..", "parsingTest");
+    Path dialectPath = Paths.get(rootPath.toAbsolutePath().toString(), "intermediate", "testDialect");
```
lisayin, line 48: nit: line a little long.

buildSrc/subprojects/parser/parser.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+
+dependencies {
+    testImplementation("org.junit.jupiter:junit-jupiter-api:5.5.1")
+    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.5.1")
+}
+
+tasks.named<Test>("test") {
+    useJUnitPlatform()
+    testLogging {
+        exceptionFormat = TestExceptionFormat.FULL
```
michaelvdow, line 31: Is this logging still needed here?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,419 @@
...
+      File currentDirectory,
+      Map<String, String> functionMap) {
+    File[] files = currentDirectory.listFiles();
+    // Ensures that files are processed first.
+    Arrays.sort(files, fileComparator);
+    String nextDirectory = directories.peek();
+    for (File f : files) {
+      String extension = "";
+      String fileName = f.getName();
+      int i = fileName.lastIndexOf('.');
```
michaelvdow, line 141: I think lines 141-145 could be simplified by using fileName.endsWith(".ftl")

buildSrc/subprojects/parser/parser.gradle.kts
AndrewPochapsky: I think it's good to have. It's like this in the main build.gradle.kts as well.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Good point, will change.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -0,0 +1,413 @@
...
+              StandardCharsets.UTF_8);
+          processFile(fileText, functionMap);
+        } catch (IOException e) {
+          e.printStackTrace();
+        } catch (IllegalStateException e) {
+          e.printStackTrace();
+        }
+      } else if (!directories.isEmpty() && fileName.equals(nextDirectory)) {
+        // Remove the front element in the queue, the value is referneced above with
+        // directories.peek() and is used in the next recursive call to this function .
```
michaelvdow, line 151: nit: extra space before period

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    String input = "{ //  } \n  }";
+    assertCurlyParserSucceeds(input);
+  }
+
+  @Test public void testCurlyParserParsesMultiComment() {
+    String input = "{ /*  } */  }";
+    assertCurlyParserSucceeds(input);
+  }
+
+  @Test public void testExtractFunctions() {
```
michaelvdow, line 147: It could be good to setup a test for when the input is malformed or missing a brace to show it should throw an IllegalStateException when it fails

AndrewPochapsky: The exception is currently being caught in the try-catch block so I don't think that would be possible.

michaelvdow: Ah I see. So would it just skip over that ftl file if it's malformed? Should the build just fail instead and throw the error rather than skipping over the logic if that's the case?

AndrewPochapsky: Currently what happens is it would add what was parsed so far to the function map so the resultant generated parserImpls.ftl will be wrong. I think it would make more sense to break the build there but I would like @dasch-1 to give some input.

AndrewPochapsky: Actually never-mind, you are right. If an error is encountered within a .ftl file, everything after that error will not be parsed.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+      put("foo", foo);
+      put("bar", bar);
+      put("baz", baz);
+      put("qux", qux);
+      put("quux", quux);
+    }};
+  }
+
+  private Queue<String> getTokens(String str) {
+    return new LinkedList(Arrays.asList(DialectGenerate.tokenizerPattern.split(str)));
```
dasch-1, line 106: nit: Line wrap

dasch-1: nit: Also, consider keeping tokenizerPattern private and instead providing a public getTokens() function in DialectGenerate.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -112,15 +110,16 @@ public void generateParserImpls(Map<String, String> functions) {
...
    */
   private Queue<String> getTraversalPath(File rootDirectoryFile) {
-    String dialectPath = dialectDirectory.getAbsolutePath();
-    String rootPath = rootDirectory.getAbsolutePath();
-    int rootIndex = dialectPath.indexOf(rootPath);
-    dialectPath = dialectPath.substring(rootIndex + rootPath.length() + 1);
-    return new LinkedList(Arrays.asList(dialectPath.split("/")));
+    Path dialectPath = dialectDirectory.toPath();
+    Path rootPath = rootDirectory.toPath();
+    dialectPath = dialectPath.subpath(rootPath.getNameCount(), dialectPath.getNameCount());
```
dasch-1, line 115: nit: Line wrap

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -112,15 +110,16 @@ public void generateParserImpls(Map<String, String> functions) {
...
+    Queue<String> pathElements = new LinkedList<>();
+    dialectPath.forEach(p -> pathElements.add(p.toString()));
+    return pathElements;
   }
 
   /**
-   * Traverses the determined path given by the queue. Once the queue is
+   * Traverses the determined path given by the directories queue. Once the queue is
```
dasch-1, line 122: nit: Line wrap

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -137,21 +136,20 @@ private void traverse(
...
+          processFile(fileText, functionMap);
+        } catch (IOException e) {
+          e.printStackTrace();
+        } catch (IllegalStateException e) {
           e.printStackTrace();
         }
       } else if (!directories.isEmpty() && fileName.equals(nextDirectory)) {
-        // Remove the front element in the queue, the value is used above with
-        // directories.peek().
+        // Remove the front element in the queue, the value is referneced above with
```
dasch-1, line 151: nit: "referneced" is misspelled

dasch-1: nit: Line wrap

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -137,21 +136,20 @@ private void traverse(
...
+        // directories.peek() and is used in the next recursive call to this function.
```
dasch-1, line 152: nit: Line wrap

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+public class DialectGenerateTest {
+
+  /**
+   * Retrns a DialectGenerate with root path of calcite/parsingTest and
+   * dialect path of calcite/parsingTest/intermediate/testDialect.
+   */
+  private DialectGenerate setupDialectGenerate() {
+    Path rootPath = Paths.get("..", "..", "..", "parsingTest");
+    // Adds the path /intermediate/testDialect/ to the end of rootPath.
+    Path dialectPath = rootPath.resolve(Paths.get("intermediate", "testDialect"));
```
dasch-1, line 47: nit: Line wrap

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.calcite.buildtools.parser.DialectGenerate.CurlyParser;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DialectGenerateTest {
+
+  /**
+   * Retrns a DialectGenerate with root path of calcite/parsingTest and
```
dasch-1, line 41: nit: "Retrns" is misspelled

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    Path rootPath = Paths.get("..", "..", "..", "parsingTest");
+    // Adds the path /intermediate/testDialect/ to the end of rootPath.
+    Path dialectPath = rootPath.resolve(Paths.get("intermediate", "testDialect"));
+    File rootFile = rootPath.toFile();
+    File dialectFile = dialectPath.toFile();
+    return new DialectGenerate(dialectFile, rootFile, "");
+  }
+
+  /**
+   * Returns the expected extraction results after calling DialectGenerate.extractFunctions
```
dasch-1, line 54: nit: Line wrap

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    return new LinkedList(Arrays.asList(DialectGenerate.tokenizerPattern.split(str)));
+  }
+
+  private void assertCurlyParserSucceeds(String input) {
+    CurlyParser parser = new CurlyParser();
+    Queue<String> tokens = getTokens(input);
+    boolean doneParsing = false;
+    while (!tokens.isEmpty()) {
+      if (doneParsing) {
+        assertTrue(false, "Failed to parse curly block");
```
dasch-1, line 115: Can you replace this assert and the "if (doneParsing)" with just a single line?

assertFalse(doneParsing, "Failed to parse curly block");

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    String input = "{ ' } '  }";
+    assertCurlyParserSucceeds(input);
+  }
+
+  @Test public void testCurlyParserParsesSingleComment() {
+    String input = "{ //  } \n  }";
+    assertCurlyParserSucceeds(input);
+  }
+
+  @Test public void testCurlyParserParsesMultiComment() {
```
dasch-1, line 142: Can we add CurlyParser test cases where the input is malformed (e.g. missing matching "}", too many "}", etc)?

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    String input = "{ /*  } */  }";
+    assertCurlyParserSucceeds(input);
+  }
+
+  @Test public void testExtractFunctions() {
+    DialectGenerate dialectGenerate = setupDialectGenerate();
+    Map<String, String> res = dialectGenerate.extractFunctions();
```
dasch-1, line 149: nit: Suggest using full variable name and not abbreviating: /res/result/

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    Map<String, String> expected = getExpectedExtractionResults();
+    assertEquals(expected.size(), res.size(),
+        "Resultant map size doesn't match expected map size");
+    for (String key : expected.keySet()) {
+      assertTrue(res.containsKey(key), "Resultant map doesn't contain expected key: " + key);
```
dasch-1, line 154: nit: Line wrap

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+      assertEquals(expected.get(key), res.get(key));
+    }
+  }
+}
```
dasch-1, line 158: In addition to the CurlyParser unit tests and the testExtractFunctions end-to-end test, did we want to add any unit tests for DialectGenerate's processFile and/or processFunction?

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -0,0 +1,158 @@
...
+    File rootFile = rootPath.toFile();
+    File dialectFile = dialectPath.toFile();
+    return new DialectGenerate(dialectFile, rootFile, "");
+  }
+
+  /**
+   * Returns the expected extraction results after calling DialectGenerate.extractFunctions
+   * onto the parsingTest directory structure.
+   */
+  private Map<String, String> getExpectedExtractionResults() {
```
dasch-1, line 57: The format of these test case strings is difficult to grok. And it could be hard for a future developer to add new test cases to.

Suggest putting the raw text in a separate test data file that this test reads from.

parsingTest/parserImpls.ftl
```diff
@@ -0,0 +1,38 @@
...
+}
+void qux(int arg1, int arg2) :
+{
+    String x = " } "
+}
+{
+    // Below is a }
+}
+
+
```
dasch-1, line 38: nit: Remove newline.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
AndrewPochapsky: I tried making it as simple as possible by having a new line where the function would have a new line. I'd prefer not having to deal with additional file IO but if you deem it necessary I'll try to change it. Any suggestion on the format of the test file?

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
AndrewPochapsky: Nice, very clean.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
dasch-1: Even if it takes a little extra work, it's good practice to follow common patterns. The pattern we should be following here is using a golden file. Here's some information about it:

https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/CurlyParserTest.java
```diff
@@ -0,0 +1,83 @@
...
+    String input = " { }";
+    assertThrows(AssertionError.class, () -> assertCurlyParserSucceeds(input));
+  }
+
+  @Test public void testCurlyParserThrowsExceptionLeadingNewLine() {
+    String input = "\n{ }";
+    assertThrows(AssertionError.class, () -> assertCurlyParserSucceeds(input));
+  }
+
+  @Test public void testCurlyParserThrowsExceptionClosedTooEarly() {
```
dasch-1, line 79: Thanks for adding these failure tests. Can we add the following 3 tests to round it out?

1) "{{ }"
2) "{"
3) "}"

AndrewPochapsky: Will do

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -0,0 +1,140 @@
...
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.Arrays;
+
+/** Traverses the parserImpls tree for the given dialect. Processing is done
+ * by <code> DialectGenerate </code>.
```
dasch-1, line 33: nit: indentation

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -0,0 +1,140 @@
...
+      return Boolean.compare(file1.isDirectory(), file2.isDirectory());
+    }
+  };
+
+  private final File dialectDirectory;
+  private final File rootDirectory;
+  private final String outputFile;
+  private final DialectGenerate dialectGenerate;
+
+  public DialectTraverser(File dialectDirectory, File rootDirectory, String outputFile) {
```
dasch-1, line 49: nit: Line wrap

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -29,89 +32,151 @@
...
+    int charIndex = dialectGenerate.processFunction(
+        DialectGenerate.getTokens(function), new LinkedHashMap<String,String>(),
+        0, declaration.length(), "foo");
+    assertEquals(function.length(), charIndex);
   }
 
-  /**
-   * Returns the expected extraction results after calling
-   * DialectGenerate.extractFunctions onto the parsingTest directory structure.
+  /** Each testcase has a testName.txt and testName_expected.txt file.
```
dasch-1, line 51: nit: Should this go one line below the "/**"?

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -29,89 +32,151 @@
...
+    String declaration = "void foo() :\n";
+    String function = declaration
+      + "{\n"
+      + "}";
+    assertThrows(IllegalStateException.class,
+        () -> assertFunctionIsParsed(declaration, function));
+  }
+
+  @Test public void processFileEmpty() {
+    String testName = "empty";
```
dasch-1, line 164: nit: Here and elsewhere, you can just omit the temporary variable and pass the name of the file in directly to assertFileProcessed().

dasch-1:
I just had a few nits. Overall, I'm really liking the new design and tests.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectTraverserTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.LinkedHashMap;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Assertions;
+import org.apache.calcite.buildtools.parser.DialectTraverser;
+
+public class DialectTraverserTest {
+  /**
+   * Returns a DialectGenerate with root path of calcite/parsingTest and
```
michaelvdow, line 34: nit: DialectGenerate -> DialectTraverser

parsingTest/intermediate/parserImpls.ftl
```diff
@@ -0,0 +1,40 @@
+/*
```
michaelvdow, line 1: Would it be better to move these files into something like a "test/resources" folder so there doesn't need to be a new folder on the root directory?

AndrewPochapsky: Do you mean where the test files for processFile are?

michaelvdow: Yeah, maybe a new folder in the same test folder where processFileTests is

AndrewPochapsky: I like that idea, I'll try it out!

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -48,17 +48,18 @@ private void assertFunctionIsParsed(String declaration, String function) {
...
+  /**
+   * Each testcase has a testName.txt and testName_expected.txt file.
    * This function makes sure that the testName.txt file is parsed without error
    * by checking that the contents of the modified functionMap match the contents
    * of the testName_expected.txt file.
    */
   private void assertFileProcessed(String testName) {
     DialectGenerate dialectGenerate = new DialectGenerate();
-    Path basePath = Paths.get("src", "test", "processFileTests", testName);
+    Path basePath = Paths.get("src", "test", "resources" ,"processFileTests", testName);
```
dasch-1, line 59: nit: line wrap

PR #113, opened 2020-06-27
[CALCITE] Added WITH CHECK OPTION support.

dstekol:


server/src/main/java/org/apache/calcite/sql/ddl/SqlCreateView.java
```diff
@@ -54,21 +55,30 @@
...
+  /** Creates a SqlCreateView (withCheckOption not specified). */
   SqlCreateView(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
       SqlIdentifier name, SqlNodeList columnList, SqlNode query) {
+    this(pos, createSpecifier, name, columnList, query,
+        /*withCheckOption=*/ false);
+  }
+
+  /** Creates a SqlCreateView (withCheckOption specified). */
+  SqlCreateView(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
+      SqlIdentifier name, SqlNodeList columnList, SqlNode query, boolean withCheckOption) {
```
dasch-1, line 72: nit: line wrap

server/src/main/java/org/apache/calcite/sql/ddl/SqlCreateView.java
```diff
@@ -86,6 +96,9 @@
     writer.keyword("AS");
     writer.newlineAndIndent();
     query.unparse(writer, 0, 0);
+    if (withCheckOption != null && withCheckOption.getValueAs(Boolean.class)) {
```
dasch-1, line 99: nit: We don't need to check for null since we control what value withCheckOption is set to from within this same class, and it's never being set to null.

PR #114, opened 2020-06-29
[CALCITE] Setup build.gradle tasks for dialects

michaelvdow:
Notes:
 - DialectGenerateTask is setup to run but currently isn't used since the generation is not finished. Two temporary tasks are included to copy over a parserImpls.ftl file into the build folder to test the fmpp/javacc tasks
 - build.gradle.kts is based off of the babel [build file](https://github.com/googleinterns/calcite/blob/master/babel/build.gradle.kts) with file paths changed
 - Parser.jj copied from core module into the parsing directory


parsing/dialects/default/config.fmpp
```diff
@@ -16,7 +16,7 @@
 data: {
   parser: {
     # Generated parser implementation class package and name
-    package: "org.apache.calcite.sql.parser.default",
+    package: "org.apache.calcite.sql.parser.defaultparser",
```
dasch-1, line 19: nit: Why change this from "default" to "defaultparser"?

michaelvdow: It was throwing an error because "default" is a reserved keyword in Java

dasch-1: Ah I see. Will there be confusion if we have the name of the dialect (e.g. "default") differ from the path/package name? Should we instead just name "default" and its corresponding package/path to "DefaultDialect"/"defaultdialect"?

michaelvdow: Yeah that makes sense. I can change everything to defaultdialect/DefaultDialect so the package and class names match

parsing/dialects/bigquery/build.gradle.kts
```diff
@@ -33,3 +36,47 @@ dependencies {
...
+    lookAhead.set(2)
+    val parserFile = fmppMain.map {
+        it.output.asFileTree.matching { include("**/Parser.jj") }.singleFile
+    }
+    inputFile.set(parserFile)
+    packageName.set("org.apache.calcite.sql.parser.bigquery")
+}
+
+ide {
+    fun generatedSource(javacc: TaskProvider<org.apache.calcite.buildtools.javacc.JavaCCTask>, sourceSet: String) =
```
lisayin, line 78: nit: line looks a little long, can you move sourceSet to a second line?

michaelvdow: Yeah I'll change that in the other gradle files too

PR #117, opened 2020-06-29
[CALCITE] Support for ALTER add column

dstekol:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1877,6 +1877,26 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test void testAlterAddColumnWithTableAttribute() {
+    final String sql = "alter table foo, no fallback add bar integer";
+    final String expected = "ALTER TABLE `FOO`, NO FALLBACK ADD"
+        + " (`BAR` INTEGER)";
+    sql(sql).ok(expected);
+  }
+
   @Test void testIndexWithoutName() {
```
dasch-1, line 1900: Let's add a failing test for 0 columns passed in.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1470,6 +1478,66 @@ SqlTableAttribute AlterTableAttributeOnCommit() :
     { return new SqlAlterTableAttributeOnCommit(getPos(), onCommitType); }
 }
 
+/**
+  * Parses a list of alter options (ex. ADD, DROP, RENAME) for
```
ArthurInTheShell, line 1482: nit: Javadoc style. [Google Java Style for Javadoc](https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line)

babel/src/main/java/org/apache/calcite/sql/babel/SqlAlterTable.java
```diff
@@ -54,12 +58,20 @@ public SqlAlterTable(SqlParserPos pos, String scope, SqlIdentifier tableName,
       int leftPrec, int rightPrec) {
     tableName.unparse(writer, leftPrec, rightPrec);
     if (tableAttributes != null) {
-      SqlWriter.Frame frame = writer.startList("", "");
+      SqlWriter.Frame tableAttributeFrame = writer.startList("", "");
```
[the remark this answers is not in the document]
dstekol: I thought it would be nice to disambiguate.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1470,6 +1478,66 @@ SqlTableAttribute AlterTableAttributeOnCommit() :
...
+  * Parses a single alter option (ex. ADD, DROP, RENAME) for
+  * ALTER TABLE queries.
+  * Used by {@code AlterTableOptions}.
+  */
+SqlAlterTableOption AlterTableOption() :
+{
+    final SqlAlterTableOption option;
+}
+{
+    (
```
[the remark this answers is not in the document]
dstekol: That is where the other options for RENAME, DROP will go. I suppose I could have just added it in the next PR, but I figured since I know there will be other things there, I will just do it in advance.

babel/src/main/java/org/apache/calcite/sql/babel/SqlAlterTable.java
```diff
@@ -34,18 +34,22 @@
 public class SqlAlterTable extends SqlAlter {
   final SqlIdentifier tableName;
```
ArthurInTheShell, line 35: nit: I am not sure if the default access specifier is appropriate for these final variable. You might consider using "private".

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1470,6 +1478,66 @@ SqlTableAttribute AlterTableAttributeOnCommit() :
...
+        { s = span(); }
+        ColumnWithType(columnList)
+        {
+            columns = new SqlNodeList(columnList, s.end(this));
+        }
+    |
+        columns = ExtendColumnList()
+    )
+    { return new SqlAlterTableAddColumn(columns); }
```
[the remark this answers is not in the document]
dstekol: Reasonable.

babel/src/main/java/org/apache/calcite/sql/babel/SqlAlterTableAddColumn.java
```diff
@@ -0,0 +1,47 @@
...
+  public final SqlNodeList columns;
+
+  public SqlAlterTableAddColumn(SqlNodeList columns) {
+    this.columns = columns;
+  }
+
+  @Override public void unparse(SqlWriter writer,
+      int leftPrec, int rightPrec) {
+    writer.keyword("ADD");
+    if (columns != null) {
```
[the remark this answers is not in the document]
dstekol: I will use the requireNonNull check suggested below, and remove the unparse check altogether.

babel/src/main/codegen/includes/parserImpls.ftl
dstekol: Assuming you mean the indentation level?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dstekol: Maybe it would actually make sense to add 2 test cases (one for the case without parentheses, and one with)?

babel/src/main/java/org/apache/calcite/sql/babel/SqlAlterTable.java
dstekol: Ok, I will mark them public, since this is what has been done in [similar classes in babel](https://github.com/googleinterns/calcite/blob/630d38ea02f57847699a2f2edde85566679a47b8/babel/src/main/java/org/apache/calcite/sql/babel/SqlCreateTable.java#L41).

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Yes.

babel/src/main/java/org/apache/calcite/sql/babel/SqlAlterTableAddColumns.java
```diff
@@ -20,28 +20,28 @@
...
  */
-public class SqlAlterTableAddColumn extends SqlAlterTableOption {
+public class SqlAlterTableAddColumns extends SqlAlterTableOption {
 
   public final SqlNodeList columns;
 
-  public SqlAlterTableAddColumn(SqlNodeList columns) {
-    this.columns = columns;
+  public SqlAlterTableAddColumns(SqlNodeList columns) {
+    this.columns = Objects.requireNonNull(columns);
```
dasch-1, line 34: Should we also assert that this is not an empty list using SqlNodeList.isEmptyList()?

dstekol: Sure, that makes sense. I guess the only situation where an empty list would be passed into the constructor would be if someone wanted to instantiate a SqlAlterTableAddColumns object and add columns to the list incrementally, but I don't think there is a situation where that would be necessary, so it's probably safe to assume that if the list is empty then it is invalid.

dasch-1: Actually, this made me think of another question: should we be supporting the use case of modifying "columns" after the SqlAlterTableAddColumns has been created?

Should we instead make a copy of columns in the constructor so that any modifications that happen after instantiation don't mutate this class' state?

dstekol: I have actually wondered about this as well. There are many places throughout the code (for example in SqlCreateTable, SqlExecMacro, SqlSelect) where a list is passed in and stored directly, without making a copy, and could therefore be modified later on, outside the class. We could make a copy here to ensure the internal state of SqlAlterTableAddColumns can't be changed, but this is really a much larger issue that isn't limited to this class. If this needs to be addressed, it might be worth opening a new refactoring feature (something along the lines of "ensuring object state immutability") that tackles this on a larger scale.

dasch-1: That sounds reasonable to me. I think it's probably fine to leave for now.

PR #118, opened 2020-06-30
[CALCITE] General logic/test fixes to pass tests for each dialect

michaelvdow:
Changes:
 - Added several @Tag annotations to SqlDialectParserTest to exclude tests not needed in Dialect1 since Dialect1 has different syntax for create table and create function
 - Changed default lookahead in several dialects that are able to pass tests with a lookahead of 1
 - Fixed parser error messages in dialects that have a lookahead of 2 by overriding getTester like in [babel](https://github.com/googleinterns/calcite/blob/master/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java#L217)
 - Added missing keywords/methods in several dialect config.fmpp files

The tests for each dialect are all passing when manually setting up the parserImpls.ftl files.


parsing/dialects/defaultdialect/src/test/java/org/apache/calcite/test/DefaultDialectQuidemTest.java
```diff
@@ -76,4 +78,34 @@ public static void main(String[] args) throws Exception {
...
+  }
+
+  /** Runs the dummy script "sql/dummy.iq", which is checked in empty but
+   * which you may use as scratch space during development. */
+
+  // Do not disable this test; just remember not to commit changes to dummy.iq
+  public void testSqlDummy(String path) throws Exception {
+    try (TryThreadLocal.Memo ignored = Prepare.THREAD_EXPAND.push(true)) {
+      checkRun(path);
+    }
```
AndrewPochapsky, line 109: This is a bit weird though I do see what it does. Would be possible to not commit this test at all after work is complete?

michaelvdow: Yeah it is kind of odd, I just copied it from the [CoreQuidemTest](https://github.com/googleinterns/calcite/blob/master/core/src/test/java/org/apache/calcite/test/CoreQuidemTest.java#L73). since some tests were also failing without the testSqlMisc method above. I tried removing testSqlDummy and testSqlScalar and the tests still pass so it can probably just be removed

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -15,6 +15,17 @@
 // limitations under the License.
 -->
 
+/* Extra operators */
+
+<DEFAULT, DQID, BTID> TOKEN :
```
dasch-1, line 20: @AndrewPochapsky will our DialectGenerate code handle this syntax in an .ftl file?

parsing/dialects/postgresqlBase/dialects/postgresql/src/test/java/org/apache/calcite/test/PostgreSQLParserTest.java
```diff
@@ -27,4 +32,46 @@
   @Override protected SqlParserImplFactory parserImplFactory() {
     return PostgreSQLParserImpl.FACTORY;
   }
+
+  /**
+   * PostgreSQL parser's global {@code LOOKAHEAD} is larger than the core
+   * parser's. This causes different parse error message between these two
+   * parsers. Here we define a looser error checker for PostgreSQL, so that we can
```
dasch-1, line 39: nit: Line wrap

parsing/dialects/postgresqlBase/dialects/redshift/src/test/java/org/apache/calcite/test/RedshiftParserTest.java
```diff
@@ -17,14 +17,61 @@
...
+        for (StackTraceElement stackTraceElement : stackTrace) {
+          String className = stackTraceElement.getClassName();
+          if (Objects.equals(className, RedshiftParserTest.class.getName())) {
+            return true;
+          }
+        }
+        return false;
+      }
+
+      private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
```
dasch-1, line 68: nit: Line wrap

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -156,14 +155,14 @@
   }
 
   /**
-   * This is a failure test making sure the LOOKAHEAD for WHEN clause is 2 in Babel, where
+   * This is a failure test making sure the LOOKAHEAD for WHEN clause is 2 in Dialect1, where
```
dasch-1, line 158: nit: Line wrap

parsing/dialects/dialect1/parserImpls.ftl
AndrewPochapsky: I don't think that it will in its current state as it is of a different format than the regular functions.

AndrewPochapsky: If we can generalize statements like this then I can come up with a parsing approach.

michaelvdow: We could also move this inside of Parser.jj and conditionally template it in with a variable since this is an outlier unless we wanted to add support for this inside of ftl files

dasch-1: Andrew and I spoke about this and he has a plan to handle extracting these clauses (TOKEN, SKIP, and MORE) from .ftl files.

PR #121, opened 2020-06-30
[CALCITE] Added support for INSERT with omitted values

michaelvdow:
Added support for omitted values in an INSERT statement:
`INSERT INTO <table_name> [VALUES] \(<value>?(,<value>?)*\);`

### DIFF
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -1039,29 +1039,32 @@ SqlNode LiteralRowConstructor() :
 }
 {
     <LPAREN>
+    e = LiteralRowConstructorItem()
+    { valueList.add(e); }
+    (
+        <COMMA>
+        e = LiteralRowConstructorItem()
+        { valueList.add(e); }
+    )*
+    <RPAREN>
+    {
+        return SqlStdOperatorTable.ROW.createCall(s.end(valueList),
+            valueList.toArray());
+    }
+}
+
+SqlNode LiteralRowConstructorItem() :
+{
+    SqlNode e;
+}
+{
     (
         e = AtomicRowExpression()
     |
         { e = SqlLiteral.createNull(getPos()); }
     )
     {
-        valueList.add(e);
-    }
-    (
-        <COMMA>
-        (
-            e = AtomicRowExpression()
-        |
-            { e = SqlLiteral.createNull(getPos()); }
-        )
-        {
-            valueList.add(e);
-        }
-    )*
-    <RPAREN>
-    {
-        return SqlStdOperatorTable.ROW.createCall(s.end(valueList),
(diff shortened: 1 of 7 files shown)


core/src/main/codegen/templates/Parser.jj
```diff
@@ -798,6 +798,11 @@ SqlNodeList ParenthesizedQueryOrCommaListWithDefault(
         e = OrderedQueryOrExpr(firstExprContext)
     |
         e = Default()
+<#if parser.includeInsertWithOmittedValues>
+    |
+        LOOKAHEAD({ getToken(1).kind != RPAREN })
```
dstekol, line 803: Maybe you could add a comment clarifying the LOOKAHEAD statement here?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1039,10 +1039,24 @@ SqlNode LiteralRowConstructor() :
 }
 {
     <LPAREN>
-    e = AtomicRowExpression() { valueList.add(e); }
     (
-        LOOKAHEAD(2)
-        <COMMA> e = AtomicRowExpression() { valueList.add(e); }
+        e = AtomicRowExpression()
```
dasch-1, line 1043: nit: Would it be worth breaking out this parsing logic into a helper function that's called in the initial case as well as in the repeated case below?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -1989,4 +1989,32 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test public void testInsertAllOmittedValues() {
+    final String sql = "insert into foo values (,,)";
+    final String expected = "INSERT INTO `FOO`\n"
+        + "VALUES (ROW(NULL, NULL, NULL))";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testInsertOneOmittedValueNoValueKeyword() {
```
dasch-1, line 2007: nit: change to testInsertOneOmittedValueNoValuesKeyword

core/src/main/codegen/templates/Parser.jj
michaelvdow: Yeah that's a good point, I'll add one here

babel/src/main/codegen/includes/parserImpls.ftl
michaelvdow: Yeah I think it'd be better to break it down like that

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
michaelvdow: Updated

PR #129, opened 2020-07-01
[CALCITE] GENERATED column attribute support.

dstekol:
Supports queries with the GENERATED column attribute, such as:
CREATE TABLE foo (bar INTEGER GENERATED ALWAYS AS IDENTITY (START WITH 5))


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -199,6 +199,143 @@ SqlColumnAttribute ColumnAttributeDefault() :
...
+            generateOptions);
+    }
+}
+
+/**
+ * Parses an option specified for the GENERATED column attribute, and adds it
+ * to a list.
+ */
+void ColumnAttributeGeneratedOption(
+    List<SqlColumnAttributeGeneratedOption> generateOptions) :
```
dasch-1, line 237: nit: generatedOptions to be consistent with the function name and class type? (same for local variable)

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGenerated.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Objects;
+
+/**
+ * A {@code SqlColumnAttributeGenerated} represents a column
+ * attribute specified by the GENERATED statement.
+ */
+public class SqlColumnAttributeGenerated extends SqlColumnAttribute {
+
+  public final GeneratedType generatedType;
+  public final List<SqlColumnAttributeGeneratedOption> generateOptions;
```
dasch-1, line 33: nit: generatedOptions to be consistent with the class type? (here and elsewhere in the file).

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedIncrementBy.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class SqlColumnAttributeGeneratedIncrementBy extends
+    SqlColumnAttributeGeneratedOption {
+
+  public final SqlLiteral inc;
+
+  /**
+   * Creates a {@code SqlColumnAttributeGeneratedIncrementBy}.
+   *
+   * @param inc    The amount specified in the INCREMENT BY attribute.
```
dasch-1, line 36: nit: Too many spaces after "inc"?

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedMaxValue.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Creates a {@code SqlColumnAttributeGeneratedMaxValue}.
+   *
+   * @param max     The amount specified in the MAXVALUE attribute.
+   * @param none    Whether NO MAXVALUE was specified.
+   */
+  public SqlColumnAttributeGeneratedMaxValue(SqlLiteral max, boolean none) {
+    this.max = max;
+    this.none = none;
+    assert none || max != null;
```
dasch-1, line 41: I know I recommended using "assert" in a previous review. However, after speaking with some peers, the recommendation is to just use Objects.requireNonNull() for nullness assertion and use comments to indicate other assertions, rather than use "assert" statements.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedMinValue.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Creates a {@code SqlColumnAttributeGeneratedMinValue}.
+   *
+   * @param min     The amount specified in the MINVALUE attribute.
+   * @param none    Whether NO MINVALUE was specified.
+   */
+  public SqlColumnAttributeGeneratedMinValue(SqlLiteral min, boolean none) {
+    this.min = min;
+    this.none = none;
+    assert none || min != null;
```
dasch-1, line 41: Same assert comment as above.

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedStartWith.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class SqlColumnAttributeGeneratedStartWith extends
+    SqlColumnAttributeGeneratedOption {
+
+  public final SqlLiteral start;
+
+  /**
+   * Creates a {@code SqlColumnAttributeGeneratedStartWith}.
+   *
+   * @param start    The amount specified in the START WITH attribute.
```
dasch-1, line 36: nit: Too many spaces after "start"?

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedMaxValue.java
dstekol: The issue is that here, "max" can be null, but only if "none" is true. Should I use Objects.requireNonNull() for "max", but enclose that in an if block so it only gets checked if "none" is false?

dasch-1: Ah I see. Then in this particular situation maybe we should just describe in a comment and avoid the assert and requireNonNull() altogether.

dstekol: Ok. In that case I should probably add a null check in the unparse method (to avoid possibly getting a null pointer error when calling max.unparse() ), right?

dasch-1: Yep, SGTM

babel/src/main/java/org/apache/calcite/sql/babel/SqlColumnAttributeGeneratedMinValue.java
```diff
@@ -32,20 +32,20 @@
   /**
    * Creates a {@code SqlColumnAttributeGeneratedMinValue}.
    *
-   * @param min     The amount specified in the MINVALUE attribute.
+   * @param min     The amount specified in the MINVALUE attribute. This
+   *    *             parameter should only be null when {@code none} is true.
```
dasch-1, line 36: nit: extra asterisk?

dasch-1:
Can you rebase on top of master and resolve the merge conflicts?

PR #130, opened 2020-07-01
[CALCITE] Gradle Task: Generating parserImpls.ftl

AndrewPochapsky:
- Added body of generateParserImpls()
- Added integration test for DialectTraverser's run() function
- Refactored common logic between tests into new TestUtils class
- Added token assignments to test-directory files


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -45,30 +45,24 @@ public int compare(File file1, File file2) {
 
   private final File dialectDirectory;
   private final File rootDirectory;
-  private final String outputFile;
+  private final String outputPathString;
```
dasch-1, line 48: nit: It's usually a convention to not append the data type to the name of a variable since this is implied by its data type.

So for here, suggest renaming to just "outputPath". But same advice throughout this PR.

PR #133, opened 2020-07-02
[CALCITE] Fixed path bug when running DialectTraverser from gradle task

AndrewPochapsky:
- Added a `src/resources/license.txt` to the root testing directory and root parsing directory
- Changed path logic in DialectTraverser to be relative to the root parsing directory and dialect directory


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -76,13 +76,16 @@ public ExtractedData extractData() {
...
+   * Generates the parserImpls.ftl file for the dialect. It is assumed that
+   * there exists a path src/resources/license.txt at the root parsing directory
+   * which was specified in the constructor.
    *
    * @param extractedData The extracted data to write to the output file
    */
   public void generateParserImpls(ExtractedData extractedData) {
-    Path outputFilePath = Paths.get(outputPath).toAbsolutePath();
-    Path licensePath = Paths.get("src", "resources", "license.txt");
+    Path outputFilePath = dialectDirectory.toPath().resolve(outputPath);
```
AndrewPochapsky, line 86: @michaelvdow Is this the expected behaviour? To give an example if the dialect id bigquery the outputFilePath would be `parsing/dialects/bigquery/$outputPath`

michaelvdow: Yeah that looks like the right path

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectTraverserTest.java
```diff
@@ -31,15 +31,15 @@
 
 public class DialectTraverserTest {
 
-  private static final Path outputPath = Paths.get("src", "test", "resources",
-      "build", "integrationTest", "actual.ftl");
+  private static final Path outputPath = Paths.get("build", "integrationTest",
+      "actual.ftl");
+  private static final Path rootPath = Paths.get("src", "test", "resources", "parserTest");
```
dasch-1, line 36: nit: line wrap

PR #134, opened 2020-07-06
[CALCITE] Added support for host variables

michaelvdow:
Host variables are variables prefixed with a colon (`:variableName`)

### DIFF
--- a/babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
+++ b/babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
@@ -38,5 +38,7 @@ public class SqlHostVariable extends SqlIdentifier {
   @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
     writer.print(":");
     writer.print(names.get(0));
+    // Ensures whitespace is added before a separator such as "AS"
+    writer.setNeedWhitespace(true);
   }
 }
--- a/babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
+++ b/babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -2255,4 +2255,51 @@ private void checkParseInfixCast(String sqlType) {
...
+
+  @Test public void testHostVariableExecNamedParams() {
+    final String sql = "exec foo (bar=:bar, baz=:baz, qux=:qux)";
+    final String expected = "EXECUTE `FOO` (`BAR` = :BAR, `BAZ` = :BAZ,"
+        + " `QUX` = :QUX)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testHostVariableSelect() {
+    final String sql = "select :bar from foo where a = :qux";
```
dasch-1, line 2273: nit: Can we add an alias after ":bar" just to cover that case too?

babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param pos  Parser position, must not be null
+   */
+  public SqlHostVariable(String name, SqlParserPos pos) {
+    super(name, pos);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.print(":");
+    writer.print(names.get(0));
+    // Ensures whitespace is added before a separator such as "AS"
```
AndrewPochapsky, line 41: nit: punctuation

babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param name  Name of the host variable
+   * @param pos  Parser position, must not be null
+   */
+  public SqlHostVariable(String name, SqlParserPos pos) {
+    super(name, pos);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.print(":");
+    writer.print(names.get(0));
```
AndrewPochapsky, line 40: nit: Can these two lines be combined into one?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
michaelvdow: Sure I can add that

babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
michaelvdow: Yeah good point, I updated it

babel/src/main/java/org/apache/calcite/sql/babel/SqlHostVariable.java
michaelvdow: Fixed

PR #136, opened 2020-07-06
[CALCITE] DialectGenerate now parses types that contain angle brackets.

AndrewPochapsky:
- Added parsing support for types that contain angle brackets
- Made function name regex more robust
- Made getFunctionName() public and added tests

### DIFF
--- a/buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
+++ b/buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
@@ -231,7 +231,7 @@ public class DialectGenerateTest {
     assertFunctionNameExtracted(declaration, "foo");
   }
 
-  @Test public void getFunctionNameSingleAngleBracketsCommaOptions() {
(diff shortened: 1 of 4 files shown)


buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -211,4 +220,24 @@ private void assertFileProcessed(String testName) {
       + "}";
     assertTokenAssignmentIsParsed(declaration, assignment);
   }
+
+  @Test public void getFunctionNameSingleWord() {
+    String declaration = "String foo () :";
+    assertFunctionNameExtracted(declaration, "foo");
+  }
+
+  @Test public void getFunctionNameSingleAngleBrackets() {
```
dasch-1, line 229: Should we have a test case with nested angle brackets? For example:

List<List<String>> foo ()

AndrewPochapsky: Sure, will add.

PR #137, opened 2020-07-06
[CALCITE] Allow non reserved keywords as host variable name

michaelvdow:


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1820,10 +1820,22 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
 {
     <COLON>
-    <IDENTIFIER> {
-        return new SqlHostVariable(unquotedIdentifier(), getPos());
-    }
+    (
+        <IDENTIFIER>
+        {
+            name = unquotedIdentifier();
+            pos = getPos();
```
AndrewPochapsky, line 1832: nit: Since this is present in both of the cases, can it be pulled out and get set after the parentheses?

michaelvdow: Yeah, we could get rid of the temp variable too and just call it in the constructor

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1820,10 +1820,14 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
 
 SqlHostVariable SqlHostVariable() :
 {
+    String name;
```
dasch-1, line 1823: nit: Can we make this final?

michaelvdow: Fixed

PR #154, opened 2020-07-09
[CALCITE] Merge master into dialectic_parsing

AndrewPochapsky:


parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -246,6 +246,49 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     };
   }
 
+  /** Tests parsing PostgreSQL-style "::" cast operator. */
+  @Test void testParseInfixCast()  {
```
michaelvdow, line 250: I think the InfixCast logic should be removed from Dialect1, it looks like it's only supported in PostgreSQL and Redshift dialects

AndrewPochapsky: Ok, should this test and corresponding parsing logic be moved to postgresqlbase?

michaelvdow: It looks like it's already in [postgresqlbase](https://github.com/AndrewPochapsky/calcite/blob/merge-master/parsing/dialects/postgresqlBase/src/test/java/org/apache/calcite/test/PostgreSQLBaseParserTest.java#L66). It should probably just be removed along with the InfixCast logic in the fmpp/ftl files

PR #155, opened 2020-07-09
[CALCITE] Hex character literal

ArthurInTheShell:
Add functionality to parse Hex character literal.


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1583,3 +1583,49 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+                charSetEnum = CharacterSet.LATIN;
+            } else if (charSet.equals("UNICODE")) {
+                charSetEnum = CharacterSet.UNICODE;
+            } else if (charSet.equals("GRAPHIC")) {
+               charSetEnum = CharacterSet.GRAPHIC;
+            } else if (charSet.equals("KANJISJIS")) {
+              charSetEnum = CharacterSet.KANJISJIS;
+            } else {
+              throw SqlUtil.newContextException(getPos(),
+              RESOURCE.unknownCharacterSet(charSet));
```
dasch-1, line 1612: nit: The indentation on this line looks off.

babel/src/main/java/org/apache/calcite/sql/babel/SqlHexCharStringLiteral.java
```diff
@@ -0,0 +1,97 @@
...
+      case GRAPHIC:
+        writer.keyword("GRAPHIC");
+        break;
+      case KANJISJIS:
+        writer.keyword("KANJISJIS");
+        break;
+      }
+    }
+    writer.keyword(value.toString());
+    switch (this.format) {
```
dasch-1, line 68: nit: Why use "this." prefix for format when not for "charSet"?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4313,6 +4316,11 @@ SqlNode StringLiteral() :
     // or comments may not occur between the prefix and the first quote, the
     // head fragment, with any prefix, is one token.
 
+<#if parser.allowHexCharacterLiteral>
+    hexCharLiteral = SqlHexCharStringLiteral()
+    {return hexCharLiteral;}
```
dasch-1, line 4321: nit: Spacing before/after open and close braces.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4313,6 +4316,11 @@ SqlNode StringLiteral() :
...
+    |
```
dasch-1, line 4322: nit: I think this "|" should be shifted all the way to the left.

babel/src/main/codegen/includes/parserImpls.ftl
lisayin, line 1612: nit: indent this line more.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1583,3 +1583,49 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+    final String formatString;
+    final String charSet;
+    CharacterSet charSetEnum = null;
+    final HexCharLiteralFormat format;
+}
+{
+    (
+        <PREFIXED_HEX_STRING_LITERAL>
+        {
+            charSet = SqlParserUtil
```
dasch-1, line 1599: This looks like a lot of Java logic in a parser file. Can we move it down into the constructor of SqlHexCharString, which we pass the raw charSet?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1583,3 +1583,49 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+              throw SqlUtil.newContextException(getPos(),
+              RESOURCE.unknownCharacterSet(charSet));
+            }
+        }
+    |
+        <QUOTED_HEX_STRING>
+    )
+    {
+        String[] tokens = token.image.split("'");
+        p = new NlsString(tokens[1], null, null);
```
lisayin, line 1620: Do we need to check that the length of tokens is always at least 2?

dasch-1, line 1620: Same suggestion here. Can we just pass in tokens[1] into the constructor of SqlHexCharString and let it do all of this processing there?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -2021,4 +2021,36 @@ private void checkParseInfixCast(String sqlType) {
...
+    expr(sql).ok(expected);
+  }
+
+  @Test void testHexCharLiteralOutsideRangeFails() {
+    // 'g' contains char outside hex range, it would be incorrectly parsed
+    // falling into the <PREFIXED_STRING_LITERAL> which leads to errors
+    final String sql = "^_unicode'cg'^XCF";
+    final String expected = "Unknown character set 'unicode'";
+    expr(sql).fails(expected);
+  }
```
dstekol, line 2055: As mentioned in design doc, I think it would also be good to test Hex Literals inside a query (ex. SELECT).

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1583,3 +1583,49 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+    final HexCharLiteralFormat format;
+}
+{
+    (
+        <PREFIXED_HEX_STRING_LITERAL>
+        {
+            charSet = SqlParserUtil
+                      .trim(SqlParserUtil.getCharacterSet(token.image), " ")
+                      .toUpperCase();
+            if (charSet.equals("LATIN")) {
```
dstekol, line 1602: nit: Maybe a switch statement would be more appropriate here to avoid having so many "else if"s?

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1583,3 +1583,49 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
             SqlLiteral.createBoolean(withTies, pos));
     }
 }
+
+SqlNode SqlHexCharStringLiteral() :
+{
+    final NlsString p;
+    final String formatString;
+    final String charSet;
+    CharacterSet charSetEnum = null;
```
dstekol, line 1592: nit: is there a reason to make charSetEnum non-final and initialize it to null (nothing else is initialized here)? It looks like in the logic below, either this enum is set to a value, or an error is thrown, so it seems like there is no need for this initialization, but I could be missing something.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Since the parser would already match either <PREFIXED_HEX_STRING_LITERAL> or <QUOTED_HEX_STRING> at this point, the token is guarantee to be in the format [_<characterSet>]'<hexValue>'[XC|XCV|XCF]. Therefore, we have guarantee that these string manipulation won't fail.

core/src/main/codegen/templates/Parser.jj
ArthurInTheShell: Following the [convention](https://github.com/googleinterns/calcite/blob/d76a6ac26a8bc662234e4e0b57c5d11f384d3c05/core/src/main/codegen/templates/Parser.jj#L4357) in the same method, this "|" could be at the right indentation.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ArthurInTheShell: Good point, will add a test case inside a query.

core/src/main/codegen/templates/Parser.jj
ArthurInTheShell: Added in https://github.com/googleinterns/calcite/pull/155/commits/0d4b39999eaa58e5cdf353dad88355c79e0ab304

babel/src/main/codegen/includes/parserImpls.ftl
lisayin: Can you add a comment here about that?

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: There is an alternative path of matching <QUOTED_HEX_STRING> without ever assigning charSetEnum a value. The error only got throw when <PREFIXED_HEX_STRING_LITERAL> is match and the characterSet is not any one of those four.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: This suggestion looks reasonable, but since we would need to call the superclass constructor at the first line of the constructor of SqlHexCharString, it seems like we still need a small part of java logic here. What I can do is to move most of the parsing logic into the constructor of SqlHexCharString, but there will probably some left.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Thanks for pointing out, changed at https://github.com/googleinterns/calcite/pull/155/commits/0e2e65e6fa9d39fa4c0b1e62109b8a8ab8f2e15c

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Thank you for pointing it out. Due to a larger refactoring at https://github.com/googleinterns/calcite/pull/155/commits/eb9c2fd414f98e835b3ef2d007f0093d7df4e3d9, this issue is outdated. The indentation level in the new file is corrected.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Sure, will add a piece of comment to clarify.

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Thank you for pointing it out. Due to a larger refactoring at https://github.com/googleinterns/calcite/pull/155/commits/eb9c2fd414f98e835b3ef2d007f0093d7df4e3d9, this issue is outdated. The indentation level in the new file is corrected.

babel/src/main/java/org/apache/calcite/sql/babel/SqlHexCharStringLiteral.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/155/commits/eb9c2fd414f98e835b3ef2d007f0093d7df4e3d9

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Refactoring done in https://github.com/googleinterns/calcite/pull/155/commits/eb9c2fd414f98e835b3ef2d007f0093d7df4e3d9 to address this issue

babel/src/main/codegen/includes/parserImpls.ftl
ArthurInTheShell: Refacotring done in https://github.com/googleinterns/calcite/pull/155/commits/eb9c2fd414f98e835b3ef2d007f0093d7df4e3d9 to address this issue.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -1586,46 +1586,27 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
-        if (formatString.equals("XC")) {
-            format = HexCharLiteralFormat.XC;
-        } else if (formatString.equals("XCV")) {
-            format = HexCharLiteralFormat.XCV;
-        } else {
-            format = HexCharLiteralFormat.XCF;
-        }
-        return new SqlHexCharStringLiteral(p, getPos(), charSetEnum, format);
+        return new SqlHexCharStringLiteral(hex, getPos(), charSet,
+        formatString);
```
dasch-1, line 1610: nit: Looks like this needs more indentation.

ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/155/commits/5cb9a502550b70ab3bc5a70a741cb4b0c645e14d

babel/src/main/java/org/apache/calcite/sql/babel/SqlHexCharStringLiteral.java
```diff
@@ -0,0 +1,136 @@
...
+        break;
+      case GRAPHIC:
+        writer.keyword("GRAPHIC");
+        break;
+      case KANJISJIS:
+        writer.keyword("KANJISJIS");
+        break;
+      }
+    }
+    writer.keyword(value.toString());
```
dstekol, line 106: nit: I think a better method to use here might be writer.literal(). This is what is used elsewhere to unparse values that are not actually keywords:

https://github.com/googleinterns/calcite/blob/5d373516fe48ad3a8c97a7c69a90bca20a4e480a/core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java#L77

https://github.com/googleinterns/calcite/blob/5d373516fe48ad3a8c97a7c69a90bca20a4e480a/core/src/main/java/org/apache/calcite/sql/SqlLiteral.java#L722

ArthurInTheShell: Fix in https://github.com/googleinterns/calcite/pull/155/commits/6a66d30df6c0e6d84e76befa55137ee96d66de4f

PR #161, opened 2020-07-10
[CALCITE] Merge dialectic_parsing into master

michaelvdow:


core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
```diff
@@ -77,11 +77,33 @@
...
+
+  /** Conformance value that instructs Calcite to use SQL semantics
+   * consistent with Redshift. */
+  REDSHIFT,
+
+  /** Conformance value that instructs Calcite to use SQL semantics
+   * consistent with MySQL version 8.x. */
+  MYSQL_8;
+
+  // TODO: change to isLeftSemiJoinAllowed, only place it's used. Remove BABEL.
```
AndrewPochapsky, line 102: nit: Are we going to deal with this TODO?

michaelvdow: I forgot about that, did we want to remove the BABEL SqlConformanceEnum?

dasch-1: Yes, let's remove it.

dasch-1: In fact, we should double check that there are no references to "server" or "babel" throughout the entire codebase.

michaelvdow: In places like [this](https://github.com/googleinterns/calcite/blob/master/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java#L481), should we use like the LENIENT conformance enum?

dasch-1: Yeah, let's use LENIENT.

PR #162, opened 2020-07-10
[CALCITE] Add support for JSON type

AndrewPochapsky:
- Moved CharacterSet enum into its own file
    - Updated KANJI -> KANJI1 (was a typo from previous feature)
- Created class SqlJsonTypeNameSpec
-  Added tests to Dialect1ParserTest
- Added parsing logic to Dialect1's parserImpls.ftl


core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
```diff
@@ -0,0 +1,117 @@
...
+      CharacterSet characterSet,
+      StorageFormat storageFormat) {
+    super(new SqlIdentifier("JSON", pos), pos);
+    this.maxLength = maxLength;
+    this.inlineLength = inlineLength;
+    this.characterSet = characterSet;
+    this.storageFormat = storageFormat;
+
+    // These two fields are mutally exclusive.
+    assert characterSet == null || storageFormat == null;
```
dasch-1, line 46: Per feedback we've gotten from the rest of the team on previous asserts, they would prefer we just capture this requirement in comments rather than have the asserts.

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
```diff
@@ -0,0 +1,117 @@
...
+
+    if (characterSet != null) {
+      assert characterSet == CharacterSet.LATIN
```
dasch-1, line 49: Per feedback we've gotten from the rest of the team on previous asserts, they would prefer we just capture this requirement in comments rather than have the asserts.

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
```diff
@@ -0,0 +1,117 @@
...
+      writer.keyword("INLINE LENGTH " + inlineLength);
+    }
+    if (characterSet != null) {
+      writer.keyword("CHARACTER SET " + characterSet.toString());
+    } else if (storageFormat != null) {
+      writer.keyword("STORAGE FORMAT " + storageFormat.toString());
+    }
+  }
+
+  public Integer getMaxLength() {
```
dasch-1, line 90: Per feedback I've received from Julian on the preferred convention in Calcite, let's convert this (and all other getters below) into just public final fields.

parsing/dialects/dialect1/config.fmpp
```diff
@@ -21,6 +21,8 @@ data: {
 
     # List of import statements.
     imports: [
+      "org.apache.calcite.sql.SqlJsonTypeNameSpec",
```
dasch-1, line 24: nit: Let's put in alphabetical order. Actually, can you reorder existing imports in this file alphabetically as well while you're editing this?

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
dasch-1: Actually, better yet, can we throw an existing exception type? They may prefer that over raw asserts and then we can enforce these constraints better than just comments. Something like:

throw SqlUtil.newContextException()

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
```diff
@@ -0,0 +1,108 @@
...
+  @Override public RelDataType deriveType(SqlValidator validator) {
+    return validator.getValidatedNodeType(getTypeName());
+  }
+
+  @Override public boolean equalsDeep(SqlTypeNameSpec spec, Litmus litmus) {
+    if (!(spec instanceof SqlJsonTypeNameSpec)) {
+      return litmus.fail("{} != {}", this, spec);
+    }
+    SqlJsonTypeNameSpec that = (SqlJsonTypeNameSpec) spec;
+    if (!this.maxLength.equals(that.maxLength)
```
ArthurInTheShell, line 70: Since this method will probably not be called by any of our existing test cases, will there be edge cases where either this.maxLength or this.inlineLength is null? We might want to add null check for this.maxLength and this.inlineLength

AndrewPochapsky: Good idea, I forgot about that case. I'll add some null checking.

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
```diff
@@ -0,0 +1,119 @@
...
+  public final Integer maxLength;
+  public final Integer inlineLength;
+  public final CharacterSet characterSet;
+  public final StorageFormat storageFormat;
+
+  public SqlJsonTypeNameSpec(SqlParserPos pos,
+      Integer maxLength,
+      Integer inlineLength,
+      CharacterSet characterSet,
+      StorageFormat storageFormat) {
```
lisayin, line 40: nit: indentation looks a little off?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1806,6 +1806,77 @@ SqlHostVariable SqlHostVariable() :
...
+}
+{
+    <JSON>
+    [
+        <LPAREN>
+        tempLiteral = UnsignedNumericLiteral() {
+            maxLength = tempLiteral.getValueAs(Integer.class);
+            if (maxLength < 2) {
+                throw SqlUtil.newContextException(getPos(),
+                    RESOURCE.numberLiteralOutOfRange(String.valueOf(maxLength)));
```
lisayin, line 1825: is this exception tested?

core/src/main/java/org/apache/calcite/sql/SqlJsonTypeNameSpec.java
AndrewPochapsky: It seems to looks consistent with other classes that I looked at such as [SqlCreateTable](https://github.com/googleinterns/calcite/blob/16820a37f33870fdc36d37e8683c511d01e30207/core/src/main/java/org/apache/calcite/sql/SqlCreateTable.java#L83). How do you think it should look like?

parsing/dialects/dialect1/config.fmpp
```diff
@@ -53,10 +52,33 @@ data: {
       "org.apache.calcite.sql.SqlColumnAttributeGeneratedMinValue",
       "org.apache.calcite.sql.SqlColumnAttributeGeneratedOption",
       "org.apache.calcite.sql.SqlColumnAttributeGeneratedStartWith",
+      "org.apache.calcite.sql.SqlColumnAttributeUpperCase",
+      "org.apache.calcite.sql.SqlCreate"
```
dasch-1, line 56: Looks like this line was accidentally added in the merge.

AndrewPochapsky: Ok, I'll get rid of it

PR #163, opened 2020-07-10
[CALCITE] Add support for BLOB data type

michaelvdow:
- Added support for the following syntax: `(BLOB | BINARY LARGE OBJECT) [ \(n[K|M|G]\) ] [(FORMAT <format_string> | TITLE <title_string> | NOT NULL)*]`
 - Created SqlColumnAttributeTitle to parse TITLE attribute
 - LobUnitSize() and SqlLobUnitSize created because these will be reused for the CLOB data type


core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * A SQL type name specification for the BLOB data type.
+ */
+public class SqlBlobTypeNameSpec extends SqlTypeNameSpec {
+
+  private SqlLiteral maxLength;
+  private SqlLobUnitSize unitSize;
+
+  /**
+   * Create a SqlBlobTypeNameSpec instance.
```
dasch-1, line 35: nit: Use "@code"

core/src/main/java/org/apache/calcite/sql/SqlColumnAttributeTitle.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * A {@code SqlColumnAttributeTitle} is the TITLE column attribute for a data type.
+ */
+public class SqlColumnAttributeTitle extends SqlColumnAttribute {
+
+  private final SqlNode titleString;
```
dasch-1, line 26: Let's make this public.

core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.calcite.util.Litmus;
+
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * A SQL type name specification for the BLOB data type.
+ */
+public class SqlBlobTypeNameSpec extends SqlTypeNameSpec {
+
+  private SqlLiteral maxLength;
```
dasch-1, line 31: Let's make this and unitSize public.

core/src/main/java/org/apache/calcite/sql/SqlColumnAttributeTitle.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * A {@code SqlColumnAttributeTitle} is the TITLE column attribute for a data type.
```
dasch-1, line 22: nit: Line wrap

parsing/dialects/dialect1/config.fmpp
```diff
@@ -99,6 +100,8 @@ data: {
       "org.apache.calcite.sql.SqlTranslateUsingCharacterSet",
       "org.apache.calcite.sql.SqlColumnAttributeDateFormat",
       "org.apache.calcite.sql.SqlHostVariable",
+      "org.apache.calcite.sql.SqlBlobTypeNameSpec",
```
dasch-1, line 103: nit: Can you put the imports in this file in alphabetical order?

parsing/dialects/dialect1/config.fmpp
```diff
@@ -1010,6 +1015,7 @@ data: {
     # Example: SqlParseTimeStampZ().
     dataTypeParserMethods: [
       "ByteIntType()"
+      "BlobDataType()"
```
dasch-1, line 1018: nit: Let's put these in alphabetical order unless we feel there's a strong reason to try one before the other during parsing.

core/src/main/java/org/apache/calcite/sql/SqlColumnAttributeTitle.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public SqlColumnAttributeTitle(SqlParserPos pos, SqlNode titleString) {
+    super(pos);
+    this.titleString = titleString;
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    writer.keyword("TITLE");
+    titleString.unparse(writer, 0, 0);
```
dstekol, line 43: May want to add a null check for titleString either here or in the constructor.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -151,7 +153,7 @@
   public static final List<SqlTypeName> ALL_TYPES =
       ImmutableList.of(
           BOOLEAN, INTEGER, VARCHAR, DATE, TIME, TIMESTAMP, NULL, DECIMAL,
-          ANY, CHAR, BINARY, VARBINARY, VARBYTE, BYTE, TINYINT, SMALLINT, BIGINT, BYTEINT, REAL,
+          ANY, CHAR, BINARY, VARBINARY, VARBYTE, BYTE, BLOB, TINYINT, SMALLINT, BIGINT, BYTEINT, REAL,
```
dstekol, line 156: nit: line is longer than 100 chars, looks like stylecheck tests are failing because of this.

core/src/main/java/org/apache/calcite/sql/SqlColumnAttributeTitle.java
michaelvdow: Sure I could add a Objects.requireNonNull call in the constructor

core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
```diff
@@ -0,0 +1,109 @@
...
+        RESOURCE.numberLiteralOutOfRange(String.valueOf(maxLength)));
+    }
+  }
+
+  private boolean isValidMaxLength(SqlLiteral maxLength,
+      SqlLobUnitSize unitSize) {
+    int numericMaxLength = maxLength.getValueAs(Integer.class);
+    switch (unitSize) {
+    case UNSPECIFIED:
+      if (numericMaxLength > 2097088000) {
```
lisayin, line 57: Add a comment on what the assumed unitSize is if it is unspecified?

michaelvdow: Sure, I could add a comment in the enum for unspecified saying it would default to bytes

core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
```diff
@@ -0,0 +1,109 @@
...
+    super(new SqlIdentifier("BLOB", pos), pos);
+    this.maxLength = maxLength;
+    this.unitSize = unitSize;
+    if (maxLength != null && !isValidMaxLength(maxLength, unitSize)) {
+      throw SqlUtil.newContextException(maxLength.getParserPosition(),
+        RESOURCE.numberLiteralOutOfRange(String.valueOf(maxLength)));
+    }
+  }
+
+  private boolean isValidMaxLength(SqlLiteral maxLength,
```
lisayin, line 52: Is maxlength == 0 a valid maxlength?

michaelvdow: I think zero would be invalid so I could add another if statement here to catch that

core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
```diff
@@ -0,0 +1,112 @@
...
+      maxLength.unparse(writer, 0, 0);
+      if (unitSize != SqlLobUnitSize.UNSPECIFIED) {
+        writer.setNeedWhitespace(false);
+        writer.print(unitSize.toString());
+      }
+      writer.sep(")");
+    }
+  }
+
+  @Override public boolean equalsDeep(SqlTypeNameSpec spec, Litmus litmus) {
```
dasch-1, line 101: What happens if we just replace all of this logic with return false? Do any tests fail?

If not, I'd suggest just doing that. It's usually better to have no logic at all than logic that's untested.

michaelvdow: Yeah it still passes all the tests, I'll change it to just false then

core/src/main/java/org/apache/calcite/sql/SqlBlobTypeNameSpec.java
lisayin: I think you're still missing this comment?

michaelvdow: I have it clarified in the [enum](https://github.com/googleinterns/calcite/blob/bde051697145e5651985547fb7ea78fb98694276/core/src/main/java/org/apache/calcite/sql/SqlLobUnitSize.java#L36), but if you you thinks it's needed I could also comment that here

PR #167, opened 2020-07-13
[CALCITE] Add support for inline CASESPECIFIC

AndrewPochapsky:
- Added java class for unparsing
- Added two functions for parsing in dialect1's parserImpls (can't combine them due to lookahead issues)
- Added tests for dialect1

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
@@ -60,7 +60,7 @@ public class SqlInlineCaseSpecific extends SqlCall
   @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
     value.unparse(writer, leftPrec, rightPrec);
     final SqlWriter.Frame frame =
-        writer.startList(SqlWriter.FrameTypeEnum.FUN_CALL, "(", ")");
+        writer.startList(SqlWriter.FrameTypeEnum.SIMPLE, "(", ")");
     if (not) {
       writer.keyword("NOT");
     }
--- a/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
@@ -40,7 +40,7 @@ public class SqlInlineCaseSpecific extends SqlCall
    *
    * @param pos  Parser position, must not be null
    * @param not  Specifies if NOT was parsed before CASE SPECIFIC
-   * @param value  The value preceding
+   * @param value  The value preceding ([NOT] CASEPECIFIC)
    */
   public SqlInlineCaseSpecific(SqlParserPos pos, boolean not, SqlNode value) {
     super(pos);
--- a/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
@@ -16,9 +16,12 @@
  */
 package org.apache.calcite.sql;
 
+import org.apache.calcite.jdbc.CalcitePrepare;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
 
 import java.util.Objects;
+import java.util.List;
 
 /**
  * Parse tree for {@code SqlInlineCaseSpecific} statement.
@@ -61,7 +64,7 @@ public class SqlInlineCaseSpecific extends SqlCall
     if (not) {
       writer.keyword("NOT");
     }
-    writer.keyword("CASE SPECIFIC");
+    writer.keyword("CASESPECIFIC");
     writer.endList(frame);
   }
 
--- a/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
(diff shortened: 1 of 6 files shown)


parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1792,6 +1792,47 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+{
+    (
+        value = StringLiteral()
+    |
+        LOOKAHEAD( CompoundIdentifier() <LPAREN> [ <NOT> ] <CASESPECIFIC> <RPAREN> )
+        value = CompoundIdentifier()
+    )
+    <LPAREN>
+    [ <NOT> { not = true; } ]
+    <CASESPECIFIC>
```
michaelvdow, line 1809: It looks like in the docs that CASESPECIFIC can also be abbreviated to CS which could be good to add in for completeness.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1792,6 +1792,47 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
     }
 }
 
+SqlNode InlineCaseSpecific() :
+{
+    SqlNode value;
```
michaelvdow, line 1797: nit: Could this value be final? Same for the below function

parsing/dialects/dialect1/parserImpls.ftl
AndrewPochapsky: Nice catch, will add.

core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.apache.calcite.jdbc.CalcitePrepare;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code SqlInlineCaseSpecific} statement.
```
dasch-1, line 27: nit: What do you think about just calling this SqlCaseSpecific?

core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class SqlInlineCaseSpecific extends SqlCall
+    implements SqlExecutableStatement {
+  public static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("INLINE_CASE_SPECIFIC",
+          SqlKind.INLINE_CASE_SPECIFIC);
```
dasch-1, line 33: nit: Same suggestion as above - just CASE_SPECIFIC.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1792,6 +1792,55 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+SqlNode InlineCaseSpecific() :
+{
+    final SqlNode value;
+    boolean not = false;
+}
+{
+    (
+        value = StringLiteral()
+    |
+        LOOKAHEAD( CompoundIdentifier() <LPAREN> [ <NOT> ] (<CASESPECIFIC>|<CS>) <RPAREN> )
```
dasch-1, line 1804: Can we put this parsing logic in a helper function (e.g. called "CaseSpecific") that you can call both from the LOOKAHEAD and below?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1792,6 +1792,55 @@ SqlNode SqlSelectTopN(SqlParserPos pos) :
...
+   and NamedFunctionCall() are options.
+ */
+SqlNode InlineCaseSpecificNamedFunctionCall() :
+{
+    final SqlNode value;
+    boolean not = false;
+}
+{
+    value = NamedFunctionCall()
+    <LPAREN>
```
dasch-1, line 1831: From line 1831 down to 1841 can all go into the helper function mentioned above (e.g. called "CaseSpecific").

parsing/dialects/dialect1/parserImpls.ftl
AndrewPochapsky: Good idea, will do.

core/src/main/java/org/apache/calcite/sql/SqlInlineCaseSpecific.java
AndrewPochapsky: Sure, sounds fine to me.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1801,19 +1801,12 @@ SqlNode InlineCaseSpecific() :
...
     )
-    <LPAREN>
-    [ <NOT> { not = true; } ]
-    (
-        <CASESPECIFIC>
-    |
-        <CS>
-    )
-    <RPAREN>
+    not = CaseSpecific()
```
dasch-1, line 1807: Rather than having CaseSpecific() return a "not" (which is a little counterintuitive), what if it just returned the constructed "SqlCaseSpecific()" object?

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -2348,6 +2348,54 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
...
+  @Test public void testInlineCaseSpecificNoneEqualsCaseSpecific() {
+    final String sql = "select * from foo where a = 'Hello' (casespecific)";
+    final String expected = "SELECT *\n"
+        + "FROM `FOO`\n"
+        + "WHERE (`A` = 'Hello' (CASESPECIFIC))";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testInlineCaseSpecificNotCaseSpecificEqualsNotCaseSpecific() {
+    final String sql = "select * from foo where a (NOT CASESPECIFIC) = 'Hello' (not casespecific)";
```
lisayin, line 2368: nit: line looks a little long

PR #168, opened 2020-07-13
[CALCITE] Add support for VARBYTE and BYTE data types

michaelvdow:
VARBYTE syntax: `VARBYTE\(n\)`
BYTE syntax: `BYTE[\(n\)]`

### DIFF
--- a/parsing/dialects/dialect1/parserImpls.ftl
+++ b/parsing/dialects/dialect1/parserImpls.ftl
@@ -1805,3 +1805,48 @@ SqlHostVariable SqlHostVariable() :
     )
     { return new SqlHostVariable(name, getPos()); }
 }
+
+SqlTypeNameSpec ByteDataType() :
+{
+    final SqlTypeName sqlTypeName;
+    final Span s = Span.of();
+    int precision = -1;
+}
+{
+    <BYTE> { sqlTypeName = SqlTypeName.BYTE; }
+    [ precision = VariableBinaryTypePrecision() ]
+    {
+        return new SqlBasicTypeNameSpec(sqlTypeName, precision, s.end(this));
+    }
+}
+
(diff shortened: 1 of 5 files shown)


parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1805,3 +1805,48 @@ SqlHostVariable SqlHostVariable() :
...
 }
+
+SqlTypeNameSpec ByteDataType() :
+{
+    final SqlTypeName sqlTypeName;
+    final Span s = Span.of();
+    int precision = -1;
+}
+{
+    <BYTE> { sqlTypeName = SqlTypeName.BYTE; }
```
dasch-1, line 1816: nit: We can probably drop the sqlTypeName variable altogether and just plug SqlTypeName.BYTE directly into the call to "new SqlBasicTypeNameSpec" below.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1805,3 +1805,48 @@ SqlHostVariable SqlHostVariable() :
...
+}
+
+SqlTypeNameSpec VarbyteDataType() :
+{
+    final SqlTypeName sqlTypeName;
+    final Span s = Span.of();
+    final int precision;
+}
+{
+    <VARBYTE> { sqlTypeName = SqlTypeName.VARBYTE; }
```
dasch-1, line 1830: nit: We can probably drop the sqlTypeName variable altogether and just plug SqlTypeName.VARBYTE directly into the call to "new SqlBasicTypeNameSpec" below.

parsing/dialects/dialect1/parserImpls.ftl
michaelvdow: Yeah that makes more sense, will update

PR #170, opened 2020-07-13
[CALCITE] Support functions in alternative CAST, inline FORMAT, and NAMED expression

michaelvdow:
This adds support for functions not wrapped in outer parentheses

### DIFF
--- a/parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
+++ b/parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
@@ -2407,4 +2407,22 @@ final class Dialect1ParserTest extends SqlDialectParserTest {
     final String expected = "SELECT CAST(`X` AS BYTEINT)";
     sql(sql).ok(expected);
   }
+
+  @Test void testAlternativeTypeConversionWithNamedFunction() {
+    final String sql = "SELECT foo(a) (INT)";
+    final String expected = "SELECT CAST(`FOO`(`A`) AS INTEGER)";
+    sql(sql).ok(expected);
+  }
+
+  @Test void testInlineFormatWithNamedFunction() {
+    final String sql = "select foo(a) (format 'X6')";
+    final String expected = "SELECT (`FOO`(`A`) (FORMAT 'X6'))";
+    sql(sql).ok(expected);
+  }
+
+  @Test void testInlinNamedWithNamedFunction() {
+    final String sql = "select foo(a) (named b)";
+    final String expected = "SELECT `FOO`(`A`) AS `B`";
+    sql(sql).ok(expected);
+  }
 }
--- a/parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
+++ b/parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
@@ -2524,7 +2524,7 @@ final class Dialect1ParserTest extends SqlDialectParserTest {
     sql(sql).ok(expected);
   }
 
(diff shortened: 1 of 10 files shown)


parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -2674,4 +2674,22 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     final String expected = "(?s).*Encountered \"\\( -\" at .*";
     sql(sql).fails(expected);
   }
+
```
ArthurInTheShell, line 2677: Just to make sure. Are we adding this feature to all parser instead of only adding it to the Dialect 1? If so, it might be more useful to put the tests in the core parser test instead.

michaelvdow: This is just for Dialect1 since postExpressionMethods is only filled in the Dialect1 config.fmpp

parsing/src/main/resources/Parser.jj
```diff
@@ -6589,9 +6590,13 @@ SqlNode NamedFunctionCall() :
             call = SqlStdOperatorTable.OVER.createCall(s.end(over), call, over);
         }
     ]
-    {
-        return call;
-    }
+    (
+<#list parser.postExpressionMethods as method>
```
ArthurInTheShell, line 6594: nit: Since this feature is for dialect 1 only, are we assuming that any parser (in the future) which has postExpressionMethods means that they would support this feature? I think we are implying whether or not to add this behavior by the existence of postExpressionMethods in a parser. I am not as familiar to this feature as you do, so could you please give me more context information on this?

michaelvdow: Sure, this is similar to how we were templating in lists of methods that only belong in babel. We would put the array, such as postExpressionMethods, in every config.fmpp file but only fill it with method names in the babel config.fmpp. Now we're just filling the array in the Dialect1 config file. If the array is empty in other dialect config files, nothing will be printed by the <#list> tag for that dialect parser.

ArthurInTheShell: Thank you for explaining it in details. What I am concern about is that if someone want to use one of the postExpressionMethods in the future in some other parser, when they add a method to the config.fmpp of that parser, this parsing behavior(not requiring parentheses for CAST/FORMAT and NAMED) will be added (unintentionally?) too.

michaelvdow: I think I see what you mean. Perhaps making a separate array for postNamedFunctionCallMethods could solve this?

ArthurInTheShell: Hmmm, I think adding another parser config boolean variable and add it to the condition statement when deciding if this feature is needed or not will help. For example, nested the <#list> </#list> statement with an <#if> statement. I don't know if I am suggesting the best way though, please try out other solution as you see fit.

michaelvdow: Yeah I think what I pushed now with an if statement works well. If some methods need to be added later for only the named function call, then another list could be added for that

parsing/src/main/resources/Parser.jj
```diff
@@ -6532,6 +6532,7 @@ SqlNode NamedFunctionCall() :
     SqlLiteral quantifier = null;
     SqlNodeList orderList = null;
     final Span withinGroupSpan;
+    SqlNode e;
```
ArthurInTheShell, line 6535: nit: Conditional injection code missing for variable SqlNode e.

PR #173, opened 2020-07-14
[CALCITE] Add support for NUMBER data type

michaelvdow:
Syntax added: `NUMBER(\((n|*)(,m)?\))?`

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
@@ -35,6 +35,12 @@ public class SqlNumberTypeNameSpec extends SqlTypeNameSpec {
   /**
    * Create a {@code SqlNumberTypeNameSpec} instance.
    *
+   * @param isPrecisionStar Whether the precision element is the symbol "*"
+   * @param precision The precision, an unsigned integer in the range [1-38].
+   *                  May be null
+   * @param scale The scale, an unsigned integer in the range [0-38] if
+   *              isPrecisionStar = true. If isPrecisionStar = false, the valid
+   *              range is [0, precision]. May be null
    * @param pos The parser position
    */
   public SqlNumberTypeNameSpec(boolean isPrecisionStar, SqlLiteral precision,
@@ -53,7 +59,8 @@ public class SqlNumberTypeNameSpec extends SqlTypeNameSpec {
     }
   }
 
-  private boolean isValidPrecision(boolean isPrecisionStar, SqlLiteral precision) {
+  private boolean isValidPrecision(boolean isPrecisionStar,
+      SqlLiteral precision) {
     if (isPrecisionStar || precision == null) {
       return true;
     }
--- a/core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
@@ -104,8 +104,8 @@ public class SqlNumberTypeNameSpec extends SqlTypeNameSpec {
     }
     SqlNumberTypeNameSpec that = (SqlNumberTypeNameSpec) spec;
     if (this.isPrecisionStar != that.isPrecisionStar
-        || this.precision != that.precision
-        || this.scale != that.scale) {
+        || this.precision.equalsDeep(that.precision, litmus)
+        || this.scale.equalsDeep(that.scale, litmus)) {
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -61,6 +61,7 @@
   FLOAT(PrecScale.NO_NO, false, Types.FLOAT, SqlTypeFamily.NUMERIC),
   REAL(PrecScale.NO_NO, false, Types.REAL, SqlTypeFamily.NUMERIC),
   DOUBLE(PrecScale.NO_NO, false, Types.DOUBLE, SqlTypeFamily.NUMERIC),
+  NUMBER(PrecScale.NO_NO, false, Types.NUMERIC, SqlTypeFamily.NUMERIC),
```
AndrewPochapsky, line 64: Is it necessary to add this here? I don't see it being referenced anywhere else in the change.

michaelvdow: Ah good catch, I was just basing it off the VARBYTE one and forgot I didn't use it here. Will remove

core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
```diff
@@ -0,0 +1,120 @@
...
+    return numericPrecision >= 1 && numericPrecision <= 38;
+  }
+
+  private boolean isValidScale(boolean isPrecisionStar, SqlLiteral precision,
+      SqlLiteral scale) {
+    if (scale == null) {
+      return true;
+    }
+    int numericScale = scale.getValueAs(Integer.class);
+    if (isPrecisionStar) {
```
dasch-1, line 77: nit: Suggest simplifying this logic slightly to:

int maxNumericPrecision = isPrecisionStar ? 38 : numericPrecision;
return numericScale >= 0 && numericScale <= maxNumericPrecision

core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  private boolean isValidPrecision(boolean isPrecisionStar,
+      SqlLiteral precision) {
+    if (isPrecisionStar || precision == null) {
+      return true;
+    }
+    int numericPrecision = precision.getValueAs(Integer.class);
+    return numericPrecision >= 1 && numericPrecision <= 38;
```
dasch-1, line 68: Let's define the lower bound and upper bounds in private static final class fields and then reuse here and elsewhere.

core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
```diff
@@ -0,0 +1,120 @@
...
+      precision.unparse(writer, 0, 0);
+    }
+    if (scale != null) {
+      writer.sep(",", true);
+      scale.unparse(writer, 0, 0);
+    }
+    writer.endList(frame);
+  }
+
+  @Override public boolean equalsDeep(SqlTypeNameSpec spec, Litmus litmus) {
```
dasch-1, line 108: What happens if we just replace all of this logic with return false? Do any tests fail?

If not, I'd suggest just doing that. It's usually better to have no logic at all than logic that's untested.

michaelvdow: Yeah it still passes all the tests, I'll change it to just false then

core/src/main/java/org/apache/calcite/sql/SqlNumberTypeNameSpec.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.calcite.util.Litmus;
+
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * A SQL type name specification for the NUMBER data type.
+ */
+public class SqlNumberTypeNameSpec extends SqlTypeNameSpec {
+
+  private final static int MIN_PRECISION = 1;
```
dasch-1, line 31: "private static final" (here and the next two lines)

michaelvdow: Whoops, got it switched. It's updated

dasch-1:
The macOS test failure says something about timing out in a JDBC unit test. I think we can safely assume that's just a test infrastructure flake.

PR #174, opened 2020-07-14
[CALCITE] New node type for EXCEPT in SELECT

AndrewPochapsky:
- Added java class SqlExcept
- Added parsing logic to BigQuery's parserImpls.ftl
- reordered imports in BigQuery's config.fmpp to be alphabetical
- Added new tests

### DIFF
--- a/core/src/main/codegen/config.fmpp
+++ b/core/src/main/codegen/config.fmpp
@@ -487,6 +487,7 @@ data: {
     includeTopN: false
     includeInsertWithOmittedValues: false
     includeHostVariables: false
+    includeExceptExpression: false
   }
 }
 
--- a/core/src/main/codegen/config.fmpp
+++ b/core/src/main/codegen/config.fmpp
@@ -469,26 +469,26 @@ data: {
     implementationFiles: [
     ]
 
+    allowAbbreviatedKeywords: false
+    allowColonSeparatorInCompoundIdentifier: false
+    allowHexCharacterStringLiteral: false
+    allowNestedJoins: false
+    allowTranslateUsingCharSetWithError: false
     allowUpdateFromTable: false
     allowUpsertFormOfUpdate: false
-    allowAbbreviatedKeywords: false
-    allowHexCharacterStringLiteral: false
-    allowTranslateUsingCharSetWithError: false
-    allowNestedJoins: false
-    includePosixOperators: false
+    includeAdditionalDeclarations: false
+    includeBraces: true
     includeCastFormat: false
     includeCompoundIdentifier: true
-    allowColonSeparatorInCompoundIdentifier: false
-    includeBraces: true
-    includeAdditionalDeclarations: false
-    includeSubstrKeyword: false
+    includeExceptExpression: false
+    includeHostVariables: false
+    includeInsertWithOmittedValues: false
+    includePosixOperators: false
+    includePostExpressionMethodsAfterFunction: false
     includeQualifyClause: false
     includeRankWithSortingExpressions: false
+    includeSubstrKeyword: false
     includeTopN: false
-    includeInsertWithOmittedValues: false
-    includeHostVariables: false
-    includeExceptExpression: false
-    includePostExpressionMethodsAfterFunction: false
   }
 }
 
--- a/core/src/main/java/org/apache/calcite/sql/SqlSelect.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlSelect.java
@@ -40,7 +40,7 @@ public class SqlSelect extends SqlCall {
 
   public SqlNodeList keywordList;
   public SqlNode topN;
-  SqlNode exceptExpression;
+  public final SqlNode exceptExpression;
   SqlNodeList selectList;
   SqlNode from;
   SqlNode where;
@@ -67,7 +67,7 @@ public class SqlSelect extends SqlCall {
       SqlNode offset,
       SqlNode fetch,
       SqlNodeList hints) {
-    this(pos, keywordList, /*topN=*/ null, /*exceptExpression=*/ null, selectList,
+    this(pos, keywordList, /*topN=*/ null, selectList, /*exceptExpression=*/ null,
         from, where, groupBy, having, /*qualify=*/ null, windowDecls, orderBy,
         offset, fetch, hints);
   }
@@ -85,7 +85,26 @@ public class SqlSelect extends SqlCall {
                    SqlNode offset,
                    SqlNode fetch,
                    SqlNodeList hints) {
-    this(pos, keywordList, /*topN=*/ null, /*exceptExpression=*/ null, selectList,
+    this(pos, keywordList, /*topN=*/ null, selectList, /*exceptExpression=*/ null,
+        from, where, groupBy, having, qualify, windowDecls, orderBy, offset,
+        fetch, hints);
+  }
+
+  public SqlSelect(SqlParserPos pos,
+      SqlNodeList keywordList,
+      SqlNode topN,
+      SqlNodeList selectList,
+      SqlNode from,
+      SqlNode where,
+      SqlNodeList groupBy,
+      SqlNode having,
+      SqlNode qualify,
+      SqlNodeList windowDecls,
+      SqlNodeList orderBy,
+      SqlNode offset,
+      SqlNode fetch,
+      SqlNodeList hints) {
+    this(pos, keywordList, topN, selectList, /*exceptExpression=*/ null,
         from, where, groupBy, having, qualify, windowDecls, orderBy, offset,
         fetch, hints);
   }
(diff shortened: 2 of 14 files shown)


core/src/main/codegen/config.fmpp
```diff
@@ -487,6 +487,7 @@ data: {
     includeTopN: false
     includeInsertWithOmittedValues: false
     includeHostVariables: false
+    includeExceptExpression: false
```
dasch-1, line 490: nit: While you're here, can you rearrange these constants so that they're in alphabetical order (same for other config.fmpp files)?

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -83,14 +85,16 @@ public SqlSelect(SqlParserPos pos,
...
                    SqlNode fetch,
                    SqlNodeList hints) {
-    this(pos, keywordList, /*topN=*/ null, selectList, from, where, groupBy,
-        having, qualify, windowDecls, orderBy, offset, fetch, hints);
+    this(pos, keywordList, /*topN=*/ null, /*exceptExpression=*/ null, selectList,
+        from, where, groupBy, having, qualify, windowDecls, orderBy, offset,
+        fetch, hints);
   }
 
   public SqlSelect(SqlParserPos pos,
```
dasch-1, line 93: Let's add a new constructor. Otherwise, we could break downstream users outside the codebase.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -262,6 +267,10 @@ public void setFetch(SqlNode fetch) {
     this.fetch = fetch;
   }
 
+  public final SqlNode getExceptExpression() {
```
dasch-1, line 270: Let's replace the getter with a public final field.

parsing/src/main/resources/Parser.jj
```diff
@@ -1333,6 +1334,11 @@ SqlSelect SqlSelect() :
         keywordList = new SqlNodeList(keywords, s.addAll(keywords).pos());
     }
     selectList = SelectList()
+<#if parser.includeExceptExpression>
+    (
```
dasch-1, line 1338: super nit: while they have the same behavior, the convention in this function appears to be to use brackets rather than parens with a "?".

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -66,8 +67,9 @@ public SqlSelect(SqlParserPos pos,
       SqlNode offset,
       SqlNode fetch,
       SqlNodeList hints) {
-    this(pos, keywordList, /*topN=*/ null, selectList, from, where, groupBy,
-        having, /*qualify=*/ null, windowDecls, orderBy, offset, fetch, hints);
+    this(pos, keywordList, /*topN=*/ null, /*exceptExpression=*/ null, selectList,
```
dasch-1, line 70: From the constructor definition, it looks like exceptExpression actually comes after selectList.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -83,14 +85,16 @@ public SqlSelect(SqlParserPos pos,
                    SqlNode offset,
                    SqlNode fetch,
                    SqlNodeList hints) {
-    this(pos, keywordList, /*topN=*/ null, selectList, from, where, groupBy,
-        having, qualify, windowDecls, orderBy, offset, fetch, hints);
+    this(pos, keywordList, /*topN=*/ null, /*exceptExpression=*/ null, selectList,
```
dasch-1, line 88: From the constructor definition, it looks like exceptExpression actually comes after selectList.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
AndrewPochapsky: Very good catch, I think that may explain my build failure as well!

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
AndrewPochapsky: I'm a bit confused what should this constructor have as inputs?

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
AndrewPochapsky: Will do, I wasn't sure which to have in this case as there looked to be a lot of existing getters of this form.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
dasch-1: One model we have been following is to have a base constructor which takes all of the fields as input. Then a variety of derived constructors that take just certain fields and then call the base constructor with the omitted fields set to null.

My comment here was that we shouldn't modify an existing derived constructor because we could then break a caller that's currently using it. Instead, we should create a brand new one which adds the exceptExpression on top of other fields.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
dasch-1: I agree, it's confusing because the original developers of Calcite didn't even follow a consistent convention themselves on this. However, Julian's recommendation to me was to use fields rather than getters, so we'll just do that going forward.

parsing/dialects/bigquery/parserImpls.ftl
```diff
@@ -14,3 +14,23 @@
...
+    final Pair<SqlNodeList, SqlNodeList> nameAndTypePair;
+    final SqlNodeList exceptList;
+}
+{
+    <EXCEPT> nameAndTypePair = ParenthesizedCompoundIdentifierList()
+    {
+        exceptList = nameAndTypePair.getKey();
+        if (selectList.size() > 0) {
+            SqlIdentifier identifier = (SqlIdentifier) selectList.get(0);
+            if (!identifier.toString().equals("*") || selectList.size() != 1) {
```
dstekol, line 29: Is it possible to have more than 1 column in the selectList and still be valid? For example, what about:
SELECT *, CURRENT_DATE as date EXCEPT(foo.a) FROM foo

AndrewPochapsky: I don't think so but all I have to go off of is the BigQuery docs which only have SELECT * EXCEPT(foo) as a valid option. What do you think @dasch-1 ?

dasch-1: Based on my reading of the BigQuery docs, only SELECT(*) EXCEPT(...) is valid.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
AndrewPochapsky: I may be missing something but I feel like this is what I did? I may be wrong but it looks like the base constructor is the one I added the field `SqlNode exceptExpression` to. And then I passed null to this field in the other existing constructors.

AndrewPochapsky: Actually nevermind, I think I see what you mean.

dasch-1: That's true, but now you'd need to add a new derived constructor to handle the previous arg list (without exceptExpression), right?

IIUC, otherwise, any callers that were using the previous version of the base constructor will automatically break after this change.

core/src/main/java/org/apache/calcite/sql/SqlSelect.java
```diff
@@ -40,7 +40,7 @@
 
   public SqlNodeList keywordList;
   public SqlNode topN;
-  SqlNode exceptExpression;
+  public final SqlNode exceptExpression;
```
dasch-1, line 43: nit: Let's keep the fields in the same order as the constructor args (e.g. exceptExpression after selectList).

core/src/main/codegen/config.fmpp
dasch-1: Thanks for doing this!

parsing/dialects/bigquery/src/test/java/org/apache/calcite/test/BigQueryParserTest.java
```diff
@@ -27,4 +29,31 @@
   @Override protected SqlParserImplFactory parserImplFactory() {
     return BigQueryParserImpl.FACTORY;
   }
+
+  @Test public void testExceptSingle() {
+    final String sql = "SELECT * EXCEPT(a) FROM foo";
+    final String expected = "SELECT * EXCEPT (`A`)\n"
+     + "FROM `FOO`";
```
dstekol, line 36: I think the indentation level of the + is causing tests to fail (tests say indentation level should be 8). Same for other tests below.

PR #177, opened 2020-07-14
[CALCITE] Add support for Sql PERIOD type

ArthurInTheShell:
By adding a SqlNameSpec type class, now user can define type to be a period type.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
@@ -59,7 +59,7 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
       }
       this.precision = precisionValue;
     } else {
-      if (timeScale == TimeScale.DATE){
+      if (timeScale == TimeScale.DATE) {
         this.precision = null;
       } else {
         this.precision = PRECISION_DEFAULT;
--- a/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.calcite.sql;
 
 import org.apache.calcite.rel.type.RelDataType;
--- a/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
@@ -32,16 +32,19 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
   public final Integer precision;
   public final boolean isWithTimezone;
 
+  private static final int PRECISION_UPPERBOUND = 6;
+  private static final int PRECISION_DEFAULT = 6;
+
   public SqlPeriodTypeNameSpec(TimeScale timeScale,
       SqlNumericLiteral precision,
       boolean isWithTimezone, SqlParserPos pos) {
     super(new SqlIdentifier("Period", pos), pos);
 
-    // date period cannot contain precision or with time zone token
+    // Date period cannot contain precision or with time zone token.
     if (timeScale == TimeScale.DATE
-        && (precision != null || isWithTimezone)) {
-      throw SqlUtil.newContextException(pos,
-          RESOURCE.illegalNonQueryExpression());
+        && precision != null || isWithTimezone) {
+          throw SqlUtil.newContextException(pos,
+              RESOURCE.illegalNonQueryExpression());
     }
 
     if (precision != null) {
@@ -50,13 +53,17 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
             RESOURCE.illegalNonQueryExpression());
       }
       int precisionValue = Integer.parseInt(precision.toValue());
-      if (precisionValue < 0 || precisionValue > 6) {
+      if (precisionValue < 0 || precisionValue > PRECISION_UPPERBOUND) {
         throw SqlUtil.newContextException(pos,
             RESOURCE.numberLiteralOutOfRange("Precision"));
       }
       this.precision = precisionValue;
     } else {
-      this.precision = null;
+      if (timeScale == TimeScale.DATE){
+        this.precision = null;
+      } else {
+        this.precision = PRECISION_DEFAULT;
+      }
     }
     this.timeScale = timeScale;
     this.isWithTimezone = isWithTimezone;
@@ -85,18 +92,7 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
   }
 
   @Override public boolean equalsDeep(final SqlTypeNameSpec spec, final Litmus litmus) {
-    if (!(spec instanceof SqlPeriodTypeNameSpec)) {
-      return litmus.fail("{} != {}", this, spec);
-    }
-
-    SqlPeriodTypeNameSpec that = (SqlPeriodTypeNameSpec) spec;
-
-    // two period types are the same as long as their time scales are the same
-    if (this.timeScale != that.timeScale) {
-      return litmus.fail("{} != {}", this, spec);
-    }
-
-    return litmus.succeed();
+    return false; // Since there is no test for this method, it is left blank.
   }
 
   public enum TimeScale {
--- a/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
@@ -28,15 +28,15 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec{
 
   public final TimeScale timeScale;
   public final Integer precision;
-  public final Boolean isWithTimezone;
+  public final boolean isWithTimezone;
 
   public SqlPeriodTypeNameSpec(TimeScale timeScale,
       SqlNumericLiteral precision,
-      Boolean isWithTimezone, SqlParserPos pos) {
+      boolean isWithTimezone, SqlParserPos pos) {
     super(new SqlIdentifier("Period",pos), pos);
 
     if (timeScale == TimeScale.DATE
-        && (precision != null || isWithTimezone != null)) {
+        && (precision != null || isWithTimezone == true)) {
       throw SqlUtil.newContextException(pos,
           RESOURCE.illegalNonQueryExpression());
     }
@@ -77,7 +77,7 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec{
       writer.endList(frame);
     }
 
-    if (isWithTimezone != null) {
+    if (isWithTimezone == true) {
       writer.keyword("WITH TIME ZONE");
     }
 
--- a/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
@@ -26,7 +26,7 @@ import static org.apache.calcite.util.Static.RESOURCE;
 /**
  * A sql type name specification of the PERIOD type.
  */
-public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec{
+public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
 
   public final TimeScale timeScale;
   public final Integer precision;
@@ -35,7 +35,7 @@ public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec{
   public SqlPeriodTypeNameSpec(TimeScale timeScale,
       SqlNumericLiteral precision,
       boolean isWithTimezone, SqlParserPos pos) {
-    super(new SqlIdentifier("Period",pos), pos);
+    super(new SqlIdentifier("Period", pos), pos);
 
     // date period cannot contain precision or with time zone token
     if (timeScale == TimeScale.DATE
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+  public final TimeScale timeScale;
+  public final Integer precision;
+  public final boolean isWithTimezone;
+
+  public SqlPeriodTypeNameSpec(TimeScale timeScale,
+      SqlNumericLiteral precision,
+      boolean isWithTimezone, SqlParserPos pos) {
+    super(new SqlIdentifier("Period",pos), pos);
+
+    // date period cannot contain precision or with time zone token
```
AndrewPochapsky, line 40: nit: punctuation.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+      writer.print(precision);
+      writer.endList(frame);
+    }
+    if (isWithTimezone) {
+      writer.keyword("WITH TIME ZONE");
+    }
+    writer.endList(periodFrame);
+  }
+
+  @Override public boolean equalsDeep(final SqlTypeNameSpec spec, final Litmus litmus) {
```
AndrewPochapsky, line 87: nit: I believe that we want to just return false in the equalsDeep functions from now on as we can't test the code.

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -2674,4 +2674,72 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     final String expected = "(?s).*Encountered \"\\( -\" at .*";
     sql(sql).fails(expected);
   }
+
+  @Test public void testPeriodTypeNameSpecsDate() {
```
AndrewPochapsky, line 2678: nit: Extra "s" after Spec (Here and also rest of the test cases)

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
dasch-1: nit: and capitalization.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+          RESOURCE.illegalNonQueryExpression());
+    }
+
+    if (precision != null) {
+      if (!precision.isInteger()) {
+        throw SqlUtil.newContextException(pos,
+            RESOURCE.illegalNonQueryExpression());
+      }
+      int precisionValue = Integer.parseInt(precision.toValue());
+      if (precisionValue < 0 || precisionValue > 6) {
```
dasch-1, line 53: nit: Let's declare the upperbound for precision into a "private static final" field.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+        throw SqlUtil.newContextException(pos,
+            RESOURCE.numberLiteralOutOfRange("Precision"));
+      }
+      this.precision = precisionValue;
+    } else {
+      this.precision = null;
```
dasch-1, line 59: nit: If I'm not mistaken, this.precision will default to null if not set.

parsing/dialects/dialect1/config.fmpp
```diff
@@ -1024,6 +1026,7 @@ data: {
     # Return type of method implementation should be "SqlTypeNameSpec".
     # Example: SqlParseTimeStampZ().
     dataTypeParserMethods: [
+      "SqlPeriodDataType()"
```
dasch-1, line 1029: nit: While you're here, can you reorder the dataTypeParserMethods to be in alphabetical order?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1999,3 +1999,35 @@ int VariableBinaryTypePrecision() :
...
+            <RPAREN>
+        ]
+        [
+            <WITH> <TIME> <ZONE> { isWithTimezone = true; }
+        ]
+    )
+    <RPAREN>
+    {
+        return new SqlPeriodTypeNameSpec(timeScale, precision, isWithTimezone,
+        getPos());
```
dasch-1, line 2031: nit: this line's indentation looks off

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Changed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Changed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
ArthurInTheShell: Changed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Changed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Good catch. It should default to 6 if not set for TIME and TIMESTAMP. Fixed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

parsing/dialects/dialect1/config.fmpp
ArthurInTheShell: Changed to alphabetical order in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

parsing/dialects/dialect1/parserImpls.ftl
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/177/commits/110d83e563426a9ea05113263c6efc02c44447dc

dasch-1:
It looks like the tests are failing with the following error:

CREATE TABLE `FOO` (`A` PERIOD(TIME WITH TIME ZONE))
    Actual:
    CREATE TABLE `FOO` (`A` PERIOD(TIME(6) WITH TIME ZONE))
    Actual java:
    "CREATE TABLE `FOO` (`A` PERIOD(TIME(6) WITH TIME ZONE))"
     ==> expected: <CREATE TABLE `FOO` (`A` PERIOD(TIME WITH TIME ZONE))> but was: <CREATE TABLE `FOO` (`A` PERIOD(TIME(6) WITH TIME ZONE))>

ArthurInTheShell:
> Failing tests.
Oh I see what is the problem with it. Will add some description to it in the design doc for discussion.

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -2695,7 +2695,7 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
 
   @Test public void testPeriodTypeNameSpecTime() {
     final String sql = "create table foo (a period(time))";
-    final String expected = "CREATE TABLE `FOO` (`A` PERIOD(TIME))";
+    final String expected = "CREATE TABLE `FOO` (`A` PERIOD(TIME(6)))";
```
dasch-1, line 2698: To be on the safe side, I'd recommend we unparse without any precision if no precision was specified in the input query.

ArthurInTheShell: Updated on https://github.com/googleinterns/calcite/pull/177/commits/07868fb1af151e9941c0be5c15dc4cb9e0d57e2f

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
dasch-1: I think my original comment here still holds: do we even need to explicitly set this.precision to null here?

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,109 @@
...
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * A sql type name specification of the PERIOD type.
+ */
+public class SqlPeriodTypeNameSpec extends SqlTypeNameSpec {
+
+  public final TimeScale timeScale;
+  public final Integer precision;
+  public final boolean isWithTimezone;
```
lisayin, line 33: nit: maybe rename to containsWithTimeZone to be more precise?

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static final int PRECISION_UPPERBOUND = 6;
+
+  public SqlPeriodTypeNameSpec(TimeScale timeScale,
+      SqlNumericLiteral precision,
+      boolean isWithTimezone, SqlParserPos pos) {
+    super(new SqlIdentifier("Period", pos), pos);
+
+    // Date period cannot contain precision or with time zone token.
```
lisayin, line 42: nit: This comment is confusing. Might be more clear if phrased something like:

TimeScale Date type cannot be used with PRECISION or WITH TIME ZONE token.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Time scale is TIME.
+     */
+    TIME,
+
+    /**
+     * Time scale is TIMESTAMP.
+     */
+    TIMESTAMP
```
lisayin, line 107: nit: These comments aren't super informative. I might leave them out or have a comment at the top of the enum describing all of them as something like Different formats for time.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: I think we need to explicitly set this.precision to null. This final variable will not be initialize in the else case if we don't do this step. If we instead set it to be null at the beginning, then this.precision can't be a final variable.

dasch-1: IIUC, this documentation suggests that we do not need to explicitly set precision to null. Instead, it will automatically inherit that value if it isn't set to something else.

https://en.wikipedia.org/wiki/Uninitialized_variable#:~:text=Java%20does%20not%20have%20uninitialized,null%20for%20all%20reference%20types).

dasch-1: If this is the case, I'd suggest just removing this explicit setting to null.

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/177/commits/e6b929deed5a3b02b9cb3368a8a219c8e4968ccd

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/177/commits/e6b929deed5a3b02b9cb3368a8a219c8e4968ccd

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/177/commits/e6b929deed5a3b02b9cb3368a8a219c8e4968ccd

parsing/dialects/dialect1/config.fmpp
dasch-1: Thanks!

core/src/main/java/org/apache/calcite/sql/SqlPeriodTypeNameSpec.java
ArthurInTheShell: See the following wiki reference on [java final variable](https://en.wikipedia.org/wiki/Final_(Java)#Final_variables).  Which it says "A blank final instance variable of a class must be definitely assigned in every constructor of the class in which it is declared;". Intellij also recognized there would be an error if I declare a final variable as a blank final but never assign it a value.

dasch-1: Ah, I didn't realize that - good to know. Ok then I agree it would be best to leave it as final and explicitly set to null in that case.

PR #178, opened 2020-07-14
[CALCITE] Added support for LIKE SOME/ANY/ALL

dstekol:
Added support for queries like:
SELECT * FROM abc WHERE a LIKE ANY(‘a’, ‘b’)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -2599,7 +2599,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   /** Returns the operator for LIKE SOME and LIKE ALL. */
   public static SqlQuantifyOperator like(SqlKind kind, boolean notLike) {
     SqlKind like = notLike ? SqlKind.NOT_LIKE : SqlKind.LIKE;
-    switch(kind) {
+    switch (kind) {
     case SOME:
     case ALL:
       return new SqlQuantifyOperator(kind, like);
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -2596,12 +2596,15 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
     }
   }
 
-  /** Returns the operator for LIKE SOME and LIKE ALL */
-  public static SqlQuantifyOperator like(SqlKind kind, boolean isLike) {
-    SqlKind like = isLike ? SqlKind.LIKE : SqlKind.NOT_LIKE;
-    if (kind == SqlKind.SOME || kind == SqlKind.ALL) {
+  /** Returns the operator for LIKE SOME and LIKE ALL. */
+  public static SqlQuantifyOperator like(SqlKind kind, boolean notLike) {
+    SqlKind like = notLike ? SqlKind.NOT_LIKE : SqlKind.LIKE;
+    switch(kind) {
+    case SOME:
+      //fall through
+    case ALL:
       return new SqlQuantifyOperator(kind, like);
-    } else {
+    default:
       throw new AssertionError(kind);
     }
   }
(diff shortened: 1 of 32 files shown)


parsing/src/main/resources/Parser.jj
```diff
@@ -3517,31 +3517,38 @@ List<Object> Expression2(ExprContext exprContext) :
                     s.clear().add(this);
                 }
                 (
+<#if parser.includeLikeAnyAllSome>
+                    LOOKAHEAD(3)
+                    LikeAnyAllSome(list, s)
+                |
+</#if>
                     (
```
dstekol, line 3525: The "changes" below this line are just an indentation shift.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2596,6 +2596,16 @@ public static SqlQuantifyOperator some(SqlKind comparisonKind) {
     }
   }
 
+  /** Returns the operator for LIKE SOME and LIKE ALL */
```
AndrewPochapsky, line 2599: nit: Punctuation

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2596,6 +2596,16 @@ public static SqlQuantifyOperator some(SqlKind comparisonKind) {
...
+  public static SqlQuantifyOperator like(SqlKind kind, boolean isLike) {
+    SqlKind like = isLike ? SqlKind.LIKE : SqlKind.NOT_LIKE;
+    if (kind == SqlKind.SOME || kind == SqlKind.ALL) {
```
dasch-1, line 2602: nit: Let's convert this to a switch/case with the default case having the AssertionError.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -2000,6 +2000,44 @@ int VariableBinaryTypePrecision() :
...
 }
 
+/**
+ * Parses LIKE ANY, LIKE ALL, and LIKE SOME statements, and appends operands
+ * to given list.
+ */
+void LikeAnyAllSome(List<Object> list, Span s) :
+{
+    final SqlOperator op;
+    boolean isLike = true;
```
dasch-1, line 2010: nit: isLike could be a little misleading since "NOT LIKE" is still a form of LIKE. Suggest changing this variable name to "not".

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -2007,18 +2007,18 @@ int VariableBinaryTypePrecision() :
...
+    boolean notLike = false;
     SqlNodeList nodeList;
 }
 {
-    [ <NOT> { isLike = false; } ]
+    [ <NOT> { notLike = true; } ]
     <LIKE>
     (
-        <SOME> { op = SqlStdOperatorTable.like(SqlKind.SOME, isLike); }
+        <SOME> { op = SqlStdOperatorTable.like(SqlKind.SOME, notLike); }
```
dasch-1, line 2017: nit: It would probably clean things up just to have a single call to SqlStdOperatorTable.like() below all of these cases. And to just set the SqlKind value within the cases.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2596,6 +2596,19 @@ public static SqlQuantifyOperator some(SqlKind comparisonKind) {
     }
   }
 
+  /** Returns the operator for LIKE SOME and LIKE ALL. */
+  public static SqlQuantifyOperator like(SqlKind kind, boolean notLike) {
+    SqlKind like = notLike ? SqlKind.NOT_LIKE : SqlKind.LIKE;
+    switch(kind) {
+    case SOME:
+      //fall through
```
lisayin, line 2604: nit: I think this is implied so you can remove this comment.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2596,6 +2596,18 @@ public static SqlQuantifyOperator some(SqlKind comparisonKind) {
     }
   }
 
+  /** Returns the operator for LIKE SOME and LIKE ALL. */
+  public static SqlQuantifyOperator like(SqlKind kind, boolean notLike) {
+    SqlKind like = notLike ? SqlKind.NOT_LIKE : SqlKind.LIKE;
+    switch(kind) {
```
dasch-1, line 2602: nit: "switch (kind) {"

PR #179, opened 2020-07-14
[CALCITE] Add support for CLOB data type

michaelvdow:
Syntax added: `(CLOB | CHARACTER LARGE OBJECT) [ \(n[K|M|G]\) ] [(CHARACTER | CHAR) SET (LATIN | UNICODE)] [(FORMAT <format_string> | TITLE <title_string> | NOT NULL)*]`


core/src/main/java/org/apache/calcite/sql/SqlClobTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+      if (characterSet == CharacterSet.LATIN && numericMaxLength > 2047937) {
+        return false;
+      } else if (characterSet == CharacterSet.UNICODE && numericMaxLength > 1023968) {
+        return false;
+      }
+      break;
+    case M:
+      if (characterSet == CharacterSet.LATIN && numericMaxLength > 1999) {
+        return false;
+      } else if (characterSet == CharacterSet.UNICODE && numericMaxLength > 999) {
```
lisayin, line 79: nit: some of these if else statements are looking a little bit long

core/src/main/java/org/apache/calcite/sql/SqlClobTypeNameSpec.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  private boolean isValidMaxLength(SqlLiteral maxLength,
+      SqlLobUnitSize unitSize, CharacterSet characterSet) {
+    int numericMaxLength = maxLength.getValueAs(Integer.class);
+    if (numericMaxLength == 0) {
+      return false;
+    }
+    switch (unitSize) {
```
lisayin, line 61: I think it might be more clear if you had two switch cases depending on the character set. Right now it's not super clear when things should return true.

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -2851,4 +2851,125 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
...
   }
 
+  @Test void testClob() {
+    final String sql = "create table foo (bar clob)";
+    final String expected = "CREATE TABLE `FOO` (`BAR` CLOB)";
+    sql(sql).ok(expected);
+  }
+
+  @Test void testCharacterLargeObject() {
+    final String sql = "create table foo (bar character large object)";
```
lisayin, line 2861: is setting a CLOB value + using `character large object` a valid use case?

core/src/main/java/org/apache/calcite/sql/SqlClobTypeNameSpec.java
michaelvdow: Yeah I could move it around so the switch case is for the character set. I'll also fix the line lengths in the above comment in this change

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
michaelvdow: Yeah, I could add a test case for that too

PR #184, opened 2020-07-15
[CALCITE] Added support for DROP MACRO

AndrewPochapsky:


core/src/main/java/org/apache/calcite/sql/SqlDropMacro.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * Parse tree for {@code DROP MACRO} statement.
+ */
+public class SqlDropMacro extends SqlDropObject {
+  private static final SqlOperator OPERATOR =
+      new SqlSpecialOperator("DROP MACRO", SqlKind.DROP_MACRO);
+
+  /** Creates a SqlDropMacro. */
```
dasch-1, line 28: nit: Use "@code"

PR #187, opened 2020-07-15
[CALCITE] Support for attributes in alternative type conversion syntax

michaelvdow:
- Data attribute support is added for alternative type conversion syntax and explicit CAST calls
 - Removed SqlFormatFunction since this functionality is now supported by the data attributes for alternative type conversion syntax


parsing/dialects/dialect1/config.fmpp
```diff
@@ -1123,15 +1124,13 @@ data: {
       "InlineCaseSpecific"
       "InlineCaseSpecificNamedFunctionCall"
       "InlineModOperator"
-      "AlternativeTypeConversionLiteralOrIdentifier"
-      "InlineFormatLiteralOrIdentifier"
       "NamedLiteralOrIdentifier"
+      "AlternativeTypeConversionLiteralOrIdentifier"
```
dasch-1, line 1128: nit: Can we order alphabetically? If not, we should probably add a comment saying why.

parsing/dialects/dialect1/config.fmpp
```diff
@@ -1123,15 +1124,13 @@ data: {
...
     ]
 
     postExpressionMethods: [
-      "AlternativeTypeConversionQuery"
-      "InlineFormatQuery"
       "NamedQuery"
+      "AlternativeTypeConversionQuery"
```
dasch-1, line 1133: nit: Can we order alphabetically? If not, we should probably add a comment saying why.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1474,50 +1486,66 @@ SqlNode AlternativeTypeConversionLiteralOrIdentifier() :
...
+            <INTERVAL> interval = IntervalQualifier() { args.add(interval); }
+        )
+        [ <COMMA> AlternativeTypeConversionAttributeList(args) ]
+    |
+        AlternativeTypeConversionAttributeList(args)
+        [
+            (
+                <COMMA> dt = DataTypeAlternativeCastSyntax() { args.add(1, dt); }
+            |
+                <COMMA> <INTERVAL> interval = IntervalQualifier() { args.add(1, interval); }
```
dasch-1, line 1542: nit: Line wrap (here and other lines in this function)

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -1746,40 +1746,40 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     sql(sql).ok(expected);
   }
 
-  @Test void testInlineFormatSimpleIdentifier() {
+  @Test void testAlternativeCastAttributeSimpleIdentifier() {
```
dasch-1, line 1749: nit: Should we have the word "format" somewhere in this test name? (Here and other functions below).

parsing/dialects/dialect1/config.fmpp
michaelvdow: There's some tests that fail if the Named method comes after. I can write a comment to explain why

parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
michaelvdow: Sure I'll update them with Format

PR #190, opened 2020-07-16
[CALCITE] Added support for RENAME MACRO statements

AndrewPochapsky:
- Added base class SqlRename to be used for RENAME nodes
   - Refactored SqlRenameTable to use this new class
   - Modified return type of RENAME parsing logic to return this class
- Added new class SqlRenameMacro
- Added parsing logic for RENAME MACRO statements in dialect1's parserImpls
- Added tests


core/src/main/java/org/apache/calcite/sql/SqlRename.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    TO,
+
+    /**
+     * Query used the AS keyword.
+     */
+    AS
+  }
+
+  private final SqlIdentifier targetStructure;
```
dasch-1, line 41: nit: Let's make these fields public so callers can more easily access them.

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1456,7 +1456,7 @@ SqlDataTypeSpec DataTypeAlternativeCastSyntax() :
     }
 }
 
-SqlRenameTable SqlRenameTable() :
+SqlRename SqlRenameTable() :
```
dasch-1, line 1459: Why not keep this return as a SqlRenameTable type?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1477,6 +1477,27 @@ SqlRenameTable SqlRenameTable() :
     }
 }
 
+SqlRename SqlRenameMacro() :
```
dasch-1, line 1480: Why not have this return the more specific SqlRenameMacro type?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -1477,6 +1477,27 @@ SqlRenameTable SqlRenameTable() :
...
+{
+    SqlIdentifier targetMacro;
```
dasch-1, line 1482: nit: Can this be final? (for this and other variables in this function).

core/src/main/java/org/apache/calcite/sql/SqlRename.java
```diff
@@ -0,0 +1,72 @@
...
+   * @param sourceStructure  What the structure is being renamed to
+   * @param renameOption  Whether the query was specified using TO or AS
+   */
+  protected SqlRename(SqlOperator operator, SqlParserPos pos,
+      SqlIdentifier targetStructure, SqlIdentifier sourceStructure,
+      RenameOption renameOption) {
+    super(operator, pos);
+    this.targetStructure = targetStructure;
+    this.sourceStructure = sourceStructure;
+    this.renameOption = renameOption;
```
dasch-1, line 59: To simplify things a little bit, do we even need to track the rename option? If there's no behavioral differences between the two cases, can we always just unparse TO and AS to the same value (e.g. AS)?

AndrewPochapsky: Not sure, I kept it his way since that is how it is done by SqlRenameTable. It would probably work to just unparse to AS each time. If so I can remove the enum completely and implement this change.

dasch-1: If it works the same, then I'd say let's go with the simpler approach of not having the enum at all.

core/src/main/java/org/apache/calcite/sql/SqlRename.java
```diff
@@ -0,0 +1,72 @@
...
+  private final SqlIdentifier targetStructure;
+  private final SqlIdentifier sourceStructure;
+  private final RenameOption renameOption;
+
+  /**
+   * Creates a {@code SqlRename}.
+   * @param operator   The specific RENAME operator
+   * @param pos  Parser position, must not be null
+   * @param targetStructure  The structure being renamed
+   * @param sourceStructure  What the structure is being renamed to
```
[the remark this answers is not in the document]
AndrewPochapsky: Interesting, I didn't know of this convention. I went off of what was in SqlRenameTable which had targetTable = table being renamed. What do you think about this @dasch-1 ?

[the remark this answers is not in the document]
dasch-1: Good point. Maybe we should just follow the nomenclature in the docs which is "old_name" and "new_name"?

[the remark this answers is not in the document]
dasch-1: If we do that, let's also change SqlRenameTable's fields to match.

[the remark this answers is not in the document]
AndrewPochapsky: That sounds good to me.

PR #192, opened 2020-07-16
[CALCITE] Added support for CREATE/REPLACE MACRO statements

AndrewPochapsky:
- Added new CREATE_MACRO frame type to ensure that SELECT statements in the body aren't unparsed with parens
- Added new AST node SqlCreateMacro and corresponding parsing logic
- Fixed bug where SELECT list created in SqlInsert was never closed

### DIFF
--- a/parsing/dialects/dialect1/parserImpls.ftl
+++ b/parsing/dialects/dialect1/parserImpls.ftl
@@ -957,6 +957,66 @@ SqlCreate SqlCreateTable() :
     }
 }
 
+SqlCreate SqlCreateMacro() :
+{
+    final Span s;
+    final SqlCreateSpecifier createSpecifier;
+    final SqlIdentifier macroName;
+    final SqlNodeList attributes;
+    final SqlNodeList sqlStatements;
+}
+{
+    (
+        <CREATE> { createSpecifier = SqlCreateSpecifier.CREATE; }
+    |
+        <REPLACE> { createSpecifier = SqlCreateSpecifier.REPLACE; }
+    )
+    {
+        s = span();
+    }
+    <MACRO> macroName = CompoundIdentifier()
+    (
+        attributes = ExtendColumnList()
+    |
+        { attributes = null; }
+    )
+    <AS>
+    sqlStatements = ParenthesizedSqlStmtList()
+    {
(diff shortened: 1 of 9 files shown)


parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -957,6 +957,67 @@ SqlCreate SqlCreateTable() :
...
+    }
+}
+
+/**
+ * Parses a list of SQL statements enclosed in parentheses and
+ * separated by semicolon. The semicolon is required between statements, but is
+ * optional at the end. Can't use existing SqlStmtList() as that function
+ * expects <EOF> at the end.
+ */
+SqlNodeList ParenthesizedSqlStmtList() :
```
dasch-1, line 997: Can we reuse SqlStmtList() instead of defining this function?

AndrewPochapsky: Unfortunately not, SqlStmtList() expects an <EOF> token after the last SqlStmt which makes it break when used in this case as the last token we have is <RPAREN> to close the macro body.

AndrewPochapsky: I tried making the <EOF> optional but that broke a few tests.

dasch-1: Ah ok, then can we factor out the common logic between the two into a helper function that both SqlStmtList() and ParenthesizedSqlStmtList() both call?

For example, we could rename the current SqlStmtList() to be SqlStmtListEof() (following the pattern of SqlStmt() vs SqlStmtEof()), move all of its logic except for the <EOF> token out into a separate function that we would be the new SqlStmtList().

Then we could have ParenthesizedSqlStmtList() call the new SqlStmtList() between a <LPAREN> and <RPAREN>.

AndrewPochapsky: Good idea, will do that.

dasch-1: Actually, we probably don't even need ParenthesizedSqlStmtList() anymore in that case and can just call SqlStmtList() between "(" and ")" tokens directly from SqlCreateMacro().

PR #194, opened 2020-07-16
[CALCITE] Added support for CREATE JOIN INDEX

michaelvdow:
- The keywords INDEX, NO, PRIMARY, and UNIQUE were removed from nonReservedKeywordsToAdd because these are reserved keywords in Dialect1. They were also causing problems with the indices list being parsed directly after the FROM clause because it would parse one of those keywords as an AS alias rather than an index.

- Added support for COLOCATE USING after MAP attribute


core/src/main/java/org/apache/calcite/sql/SqlCreateJoinIndex.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+
+/**
+ * Parse tree for {@code CREATE JOIN INDEX} statement.
+ */
+public class SqlCreateJoinIndex extends SqlCreate {
+  private static final SqlOperator OPERATOR =
+      new SqlSpecialOperator("CREATE JOIN INDEX", SqlKind.CREATE_TABLE);
```
dasch-1, line 29: Let's create a new SqlKind for CREATE_JOIN_INDEX.

core/src/main/java/org/apache/calcite/sql/SqlCreateJoinIndex.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /** Creates a {@code SqlCreateJoinIndex}. */
+  public SqlCreateJoinIndex(SqlParserPos pos, SqlIdentifier name,
+      List<SqlTableAttribute> tableAttributes, SqlNodeList select, SqlNode from,
+      SqlNode where, SqlNodeList groupBy, SqlNodeList orderBy,
+      List<SqlIndex> indices) {
+    super(OPERATOR, pos, SqlCreateSpecifier.CREATE, false);
+    this.name = name;
+    this.tableAttributes = tableAttributes;
+    this.select = select;
```
dasch-1, line 48: Sorry I didn't catch this sooner in the design doc, but rather than recreating all of the fields of the SELECT statement, can we leverage the existing SELECT parsing logic and nodes?

For example, similar to how OrderedQueryOrExpr returns a SqlNode object and sets that as the "source" in the SqlInsert() parsing method?

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -517,10 +517,12 @@ SqlTableAttribute TableAttributeJournalTable() :
 SqlTableAttribute TableAttributeMap() :
 {
     final SqlIdentifier id;
+    SqlIdentifier colocateName = null;
 }
 {
     <MAP> <EQ> id = CompoundIdentifier()
-    { return new SqlTableAttributeMap(id, getPos()); }
+    [ <COLOCATE> <USING> colocateName = SimpleIdentifier() ]
```
dasch-1, line 524: Is it ever legal for the colocateName to be a CompoundIdentifier?

core/src/main/java/org/apache/calcite/sql/SqlCreateJoinIndex.java
michaelvdow: Yeah I think that would be a lot simpler. Then I can remove a bunch of variables from SqlCreateJoinIndex and the copied methods from SqlSelect(). I'll work on refactoring this to use OrderedQueryOrExpr.

parsing/dialects/dialect1/parserImpls.ftl
michaelvdow: It looks like it is valid, it's changed to CompoundIdentifier()

parsing/dialects/dialect1/parserImpls.ftl
```diff
@@ -2259,3 +2261,51 @@ SqlDrop SqlDropMacro(Span s, boolean replace) :
...
+        (
+            e = TableAttributeMap()
+        |
+            e = TableAttributeFallback()
+        |
+            e = TableAttributeChecksum()
+        |
+            e = TableAttributeBlockCompression()
+        ) { list.add(e); }
+    )+
```
dstekol, line 2282: nit: If you changed the + above to *, I think this could enable you to simplify the logic in other places. For example, you could remove the square brackets around the call to CreateJoinIndexTableAttributes (line 2298), you could declare the tableAttributes variable (line 2290) as final and avoid initalizing it to null (if there are no attributes, it just gets set to an empty list), and then you can remove the null check for tableAttributes from the SqlCreateJoinIndex unparsing function and just do Objects.requireNonNull in the constructor.

michaelvdow: Yeah I like that, it could help remove some unnecessary checks. I'll work on updating this

core/src/main/java/org/apache/calcite/sql/SqlCreateJoinIndex.java
```diff
@@ -0,0 +1,74 @@
...
+  public final List<SqlTableAttribute> tableAttributes;
+  public final SqlNode select;
+  public final List<SqlIndex> indices;
+
+  /** Creates a {@code SqlCreateJoinIndex}. */
+  public SqlCreateJoinIndex(SqlParserPos pos, SqlIdentifier name,
+      List<SqlTableAttribute> tableAttributes, SqlNode select,
+      List<SqlIndex> indices) {
+    super(OPERATOR, pos, SqlCreateSpecifier.CREATE, false);
+    this.name = name;
```
dstekol, line 42: I think it would still make sense to add an Objects.requireNonNull() here before setting each of these fields, since even though the existing parsing logic wouldn't set them to null, it's possible some other parsing function could be created that would also instantiate this object.

michaelvdow: That's a good point. I'll add it in.

PR #197, opened 2020-07-17
[CALCITE] Fixed recursion issue for named function calls

michaelvdow:


parsing/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -3311,4 +3311,69 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     final String expected = "DROP MACRO `FOO`";
     sql(sql).ok(expected);
   }
+
+  @Test void testNestedNamedFunctionCalls() {
```
michaelvdow, line 3315: @dasch-1 I'm not sure what kind of test to run for this fix so I just included the test from the design doc for now

dasch-1: This looks good to me. Do we know approximately how long it takes to run this new test case?

michaelvdow: On my local it was running almost instantly.

PR #208, opened 2020-07-22
[CALCITE] Remove injection logic

AndrewPochapsky:


AndrewPochapsky:
> This is awesome!
> 
> One thought: should we add some sort of programmatic enforcement so that injection logic isn't mistakenly added by a programmer in the future? Could we detect any <#if> or <#list> expressions in the DialectGenerate task and fail the build if any are found?

Sounds like a good idea. Though there is still a tiny bit of injection logic for the keywords which I think is required: https://github.com/googleinterns/calcite/blob/50d6732f52d928dd6e7279c03ff1105637a57a45/parsing/parserImpls.ftl#L6862

dasch-1:
Ah ok, then we can hold off on implementing the enforcement logic while these exist.

However, ultimately, I think it would be nice to simplify the keyword injection logic as well. It would be cool if we could just statically define all keywords for a given dialect in the corresponding config.fmpp and/or parserImpls.ftl files directly.

However, we can track that in a separate feature request if you think it would require a nontrivial amount of work.

AndrewPochapsky:
> Ah ok, then we can hold off on implementing the enforcement logic while these exist.
> 
> However, ultimately, I think it would be nice to simplify the keyword injection logic as well. It would be cool if we could just statically define all keywords for a given dialect in the corresponding config.fmpp and/or parserImpls.ftl files directly.
> 
> However, we can track that in a separate feature request if you think it would require a nontrivial amount of work.

Sounds good, I think doing this in a separate feature makes sense.

PR #211, opened 2020-07-22
[CALCITE] Merging parsing_exodus into master

AndrewPochapsky:
- Created new top level parsing/parserImpls.ftl file containing logic that was previously in core
    - Moved existing parsing/parserImpls.ftl and dialects into new directory parsing/intermediate
- Moved core's config.fmpp to parsing/
- Changed path in core's build.gradle.kts to reference this config.fmpp file
- Removed all injection logic (except keywords for now) and replaced with an overriding approach


core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -255,13 +255,7 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
...
       writer.print("-");
     }
-    Long intervalValueInLong;
-    try {
-      intervalValueInLong = Long.parseLong(literal.getValue().toString());
-    } catch (NumberFormatException e) {
-      throw new RuntimeException("Only INT64 is supported as the interval value for BigQuery.");
-    }
-    writer.literal(intervalValueInLong.toString());
+    writer.literal(literal.getValue().toString());
```
dasch-1, line 258: Hm, this shouldn't be in the diff. Doesn't this change already exist in master?

parsing/dialects/bigquery/parserImpls.ftl
```diff
@@ -1,36 +0,0 @@
-<#--
```
dasch-1, line 1: Why is this file showing up as deleted rather than moved/renamed?

parsing/intermediate/dialects/postgresqlBase/dialects/postgresql/parserImpls.ftl
```diff
@@ -0,0 +1,196 @@
+<#--
```
dasch-1, line 1: Why is this file showing up as new, rather than moved/renamed?

parsing/intermediate/dialects/postgresqlBase/dialects/redshift/parserImpls.ftl
```diff
@@ -0,0 +1,196 @@
+<#--
```
dasch-1, line 1: Same question for this file on why this isn't moved/renamed?

parsing/intermediate/dialects/postgresqlBase/parserImpls.ftl
```diff
@@ -0,0 +1,335 @@
+<#--
```
dasch-1, line 1: Same question for this file on why this isn't moved/renamed?

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
AndrewPochapsky: No idea why any of what you pointed out is happening :/ Maybe because this isn't the original branch?

AndrewPochapsky: Actually the same thing is happening with parsing_exodus into master: https://github.com/googleinterns/calcite/compare/master...parsing_exodus

AndrewPochapsky: My guess is that since the /intermediate/ directory didn't exist before git considers it as a new file.

dasch-1: Interestingly, not all .ftl files are showing this issue. For example, this file looks like it was moved/renamed:

parsing/dialects/defaultdialect/parserImpls.ftl → parsing/intermediate/dialects/defaultdialect/parserImpls.ftl

AndrewPochapsky: I don't believe that I made any changes to defaultdialect which is maybe why it is showing up properly.

dasch-1: I don't think it's just a matter of which files were edited because dialect1/parserImpls.ftl also was moved/renamed properly:
parsing/dialects/dialect1/parserImpls.ftl → parsing/intermediate/dialects/dialect1/parserImpls.ftl

Can you try doing a "git mv" on the files that are showing up as deleted/created? (https://git-scm.com/docs/git-mv)

AndrewPochapsky: Sorry if I'm missing something, but how can I do this if the changes have already been made?

dasch-1: You should be able to use some combination of "git reset" and "git checkout" with a specific filename to undo the changes in that file to match the version in master (https://git-scm.com/docs/git-checkout).

Just make sure to copy/paste the contents of the file into some place safe so that you can copy/paste them back in after you do the "git mv".

dasch-1: I ran through the steps I recommended to you and I'm still seeing GitHub show these files as being deleted/added rather than modified/renamed, even though git was saying otherwise. I think we can chalk this up to a GitHub diff visualization bug.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3022,10 +3022,6 @@ private void checkLiteral2(String expression, String expected) {
             + "WHERE (hire_date + INTERVAL 10 HOUR)"
             + " > TIMESTAMP '2005-10-17 00:00:00'";
     sql(sql1).withBigQuery().ok(expect1);
-
-    final String sql2 = "select  * from \"employee\" where  \"hire_date\" + "
-            + "INTERVAL '1 2:34:56.78' DAY TO SECOND > TIMESTAMP '2005-10-17 00:00:00' ";
-    sql(sql2).withBigQuery().throws_("Only INT64 is supported as the interval value for BigQuery.");
```
ArthurInTheShell, line 3028: Could you provide more context on the deletion of this test cases?

AndrewPochapsky: That wasn't my change actually, not sure why it shows up in the diff. It was made in this Pr: https://github.com/googleinterns/calcite/pull/207

PR #213, opened 2020-07-23
[CALCITE] Added dialectic parsing test tool to create report of failing queries

michaelvdow:
Can be ran using the command: `./gradlew :parsing:tools:findDialectParsingErrors -PinputPath <input_path> -PoutputPath <outputPath> -Pdialect=<dialect> [-PgroupByErrors] [-PnumSampleQueries=<num>]` where the paths are relative to the tools directory.

This tool takes in a CSV file of queries and will output a JSON file containing a report of failing queries.


parsing/tools/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+plugins {
+    java
+    kotlin("jvm")
+}
+
+dependencies {
+    api(project(":parsing:dialects:bigquery"))
```
AndrewPochapsky, line 23: Just a note, once https://github.com/googleinterns/calcite/pull/211 gets merged these dependencies will need to be changed to :parsing:intermediate:...

michaelvdow: Thanks for the heads up, I'll change it once 211 gets merged since that'll probably be merged first.

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,402 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Processes a CSV file containing queries and will output a JSON file containing the results of
```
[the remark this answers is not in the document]
michaelvdow: Ah my editor was set to 100 chars, will update

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,402 @@
...
+  private final String inputPath;
+  private final String outputPath;
+  private final Dialect dialect;
+  private final boolean groupByErrors;
+  private final int numSampleQueries;
+  private final List<MessageFormat> errorFormats;
+
+  /**
+   *  Creates a new instance of {@code FindDialectParsingErrors}. Also populates the errorFormats
+   *  list with a MessageFormat object for each custom defined error message in CalciteResource.
```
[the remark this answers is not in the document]
michaelvdow: Updated

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,402 @@
...
+      this.count = count;
+    }
+  }
+
+  enum Dialect {
+    BIGQUERY {
+      @Override public SqlParserImplFactory getDialectFactory() {
+        return BigQueryParserImpl.FACTORY;
+      }
+    }, DEFAULTDIALECT {
```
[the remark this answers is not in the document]
michaelvdow: Updated

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+  private final String inputPath;
+  private final String outputPath;
+  private final Dialect dialect;
+  private final boolean groupByErrors;
+  private final int numSampleQueries;
+  private final List<MessageFormat> errorFormats;
+
+  /**
+   *  Creates a new instance of {@code FindDialectParsingErrors}. Also populates
+   *  the errorFormats list with a {@Code MessageFormat} object for each custom
```
dasch-1, line 70: /@Code/@code/

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+   *                         type
+   */
+  public FindDialectParsingErrors(String inputPath, String outputPath,
+      Dialect dialect, boolean groupByErrors, int numSampleQueries) {
+    this.inputPath = inputPath;
+    this.outputPath = outputPath;
+    this.dialect = dialect;
+    this.groupByErrors = groupByErrors;
+    this.numSampleQueries = numSampleQueries;
+    errorFormats = new ArrayList<>();
```
lisayin, line 91: should this be this.errorFormats?

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+  private final boolean groupByErrors;
+  private final int numSampleQueries;
+  private final List<MessageFormat> errorFormats;
+
+  /**
+   *  Creates a new instance of {@code FindDialectParsingErrors}. Also populates
+   *  the errorFormats list with a {@code MessageFormat} object for each custom
+   *  defined error message in CalciteResource.
+   *
+   * @param inputPath Path to the input CSV file containing queries
```
lisayin, line 73: maybe add: Assumes CSV file is delimited by commas.

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
dasch-1: I believe it's legal for it to be with or without the "this" prefix. I'm guessing Michael didn't use the "this" prefix in this case because it wasn't needed to disambiguate it from a local variable with the same name.

Nevertheless, it might just be clearer to use "this.errorFormats" in the constructor (but not necessarily anywhere else) since the other fields are being set this way.

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+   * the results of failing queries. This will run findErrorGroups if
+   * groupByErrors is true. Otherwise, findFullErrors will run.
+   *
+   * @throws IOException If it fails to read the input file
+   */
+  public void run() throws IOException {
+    CSVReader reader = new CSVReader(
+        new InputStreamReader(new FileInputStream(inputPath),
+        StandardCharsets.UTF_8), ',', '"', 1);
+    List<String[]> rows = reader.readAll();
```
lisayin, line 113: Will there potentially be a memory issue if the reader puts all of the queries in memory at the same time instead of streaming the queries?

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+    reader.close();
+    List<String> queries = rows.parallelStream()
```
lisayin, line 115: Add a comment about what is being parallelized here. Will the next line only run after all of the rows are processed?

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,424 @@
...
+        .collect(Collectors.toList());
+    if (groupByErrors) {
+      findErrorGroups(queries);
+    } else {
+      findFullErrors(queries);
+    }
+  }
+
+  /**
+   * This is the default processing method. It parallel processes the list of
```
lisayin, line 126: nit: It processes the list of queries in parallel....

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: Sure I can add that in for clarification

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: There might be if the file is extremely large. However, I'm not sure if I can use the paralellStream method without reading in the entire file at once. I think the only other option would be to read queries in one at a time and process them sequentially but that might slow it down.

michaelvdow: I think if I processed it sequentially it might still use the same amount of memory because it needs to store all the results since it sorts it at the end for the group by method

dasch-1: Thanks for pointing this out, Lisa! My vote would be to keep FileIO as simple as possible for now until we encounter a problem here. As is, I've probably already complicated things enough by requesting that Michael process the queries in parallel :)

The targeted use case is to run through O(thousands) of queries at a time, which *shouldn't* require too much memory. However, if we find this assumption not to be true in practice or that target use case changes, then I think it might be worth revisiting this.

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: That's a good idea, I'll add that it expects a header too since it skips the first line

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: Sure, I added a comment above it clarifying it sanitizes queries in parallel

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: Updated

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
lisayin: Maybe consider leaving a comment here about it to help pinpoint where it might cause issues in the future?

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
```diff
@@ -0,0 +1,427 @@
...
+    replaceChars.put('\u2018', '\'');
+    replaceChars.put('\u2019', '\'');
+    replaceChars.put('\u201c', '\"');
+    replaceChars.put('\u201d', '\"');
+    replaceChars.put('\u2028', ' ');
+    for (Map.Entry<Character, Character> entry : replaceChars.entrySet()) {
+      query = query.replace(entry.getKey(), entry.getValue());
+    }
+    query = query.trim();
+    if (query.endsWith(";")) {
```
lisayin, line 351: will whitespace after a semicolon be a possible in the provided queries?

michaelvdow: Yeah some of the CSV files I was given had whitespace after it, but the trim() call above it should remove whitespace on both sides of the query.

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
michaelvdow: Yeah that would be a good idea, I'll leave a comment here

parsing/tools/src/main/java/com/apache/calcite/parsing/tools/FindDialectParsingErrors.java
lisayin: oh my bad i didn't see that line

PR #215, opened 2020-07-23
[CALCITE] File annotations for DialectGenerate

AndrewPochapsky:
- Each extracted function and token assignment now has a comment above its declaration stating which file it was extracted from
- Added a new line separator between each extracted entity to improve readability
- Modified existing tests to work with annotations


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -143,9 +143,15 @@ private void traverse(
       String fileName = f.getName();
       if (f.isFile() && fileName.endsWith(".ftl")) {
         try {
-          String fileText = new String(Files.readAllBytes(f.toPath()),
+          Path absoluteFilePath = f.toPath();
```
dasch-1, line 146: nit: This is getting to be a lot of contents in this try/catch block. Can we split out some things to not be in the try/catch?

buildSrc/subprojects/parser/src/test/resources/integrationTest/expected.ftl
```diff
@@ -14,10 +14,14 @@
...
  * limitations under the License.
  */
+
+// Extracted from: parserTest/parserImpls.ftl
 <DEFAULT> TOKEN :
 {
 
 }
+
+// Extracted from: parserTest/intermediate/parserImpls.ftl
```
dasch-1, line 24: Should we have the parsingTest directory structure match the real one? e.g. parserTest/dialects/intermediate/parserImpls.ftl here?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -74,8 +74,10 @@
    * @param fileText The contents of the file to process
    * @param extractedData The object to which the parsed functions and token assignments
    *             will be added to
+   * @param filePath The path of the file being processed
    */
-  public void processFile(String fileText, ExtractedData extractedData) {
+  public void processFile(String fileText, ExtractedData extractedData,
+      String filePath) {
```
lisayin, line 80: Will there be other places that might call processFile without filePath that would still be a legitimate call?

buildSrc/subprojects/parser/src/test/resources/integrationTest/expected.ftl
AndrewPochapsky: Good idea, will make the change.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Not at the moment but I can see it possibly happening in the future. I'll make this parameter optional by having a null check before the annotation gets added.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
AndrewPochapsky: Will do

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -168,31 +169,31 @@ private void assertFileProcessed(String testName) {
   }
 
   @Test public void processFileEmpty() {
-    assertFileProcessed("empty");
+    assertFileProcessed("empty", true);
```
dasch-1, line 172: nit: Use comment (here and elsewhere) to indicate the name of the parameter (e.g. /*specifyFilePath=*/)

PR #217, opened 2020-07-24
[CALCITE] Best effort validation.

dstekol:
Added support for handling unknown table identifiers and columns deriving from unknown tables.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
@@ -241,7 +241,7 @@ public abstract class DelegatingScope implements SqlValidatorScope {
     if (filteredMap.isEmpty() && !map.isEmpty()) {
       String arbitraryKey = map.keySet().iterator().next();
       filteredMap.put(arbitraryKey, map.get(arbitraryKey));
-      //todo(dstekol): Figure out how to represent ambiguous columns
+      //TODO(dstekol): Figure out how to represent ambiguous columns
     }
     return filteredMap;
   }
--- a/core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
@@ -21,6 +21,7 @@ import org.apache.calcite.rel.type.DynamicRecordType;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rel.type.StructKind;
+import org.apache.calcite.rel.type.UnknownRecordType;
 import org.apache.calcite.schema.CustomColumnResolvingTable;
 import org.apache.calcite.schema.Table;
 import org.apache.calcite.sql.SqlCall;
@@ -39,6 +40,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -85,7 +87,7 @@ public abstract class DelegatingScope implements SqlValidatorScope {
   }
 
   public void resolve(List<String> names, SqlNameMatcher nameMatcher,
-      boolean deep, Resolved resolved) {
+                      boolean deep, Resolved resolved) {
     parent.resolve(names, nameMatcher, deep, resolved);
   }
 
@@ -212,6 +214,47 @@ public abstract class DelegatingScope implements SqlValidatorScope {
 
   public SqlValidator getValidator() {
     return validator;
+  }
+
+  /**
+   *
+   * @param map A map from table names to {@code ScopeChild} objects, which
+   *            represents all possible tables which a particular column
+   *            could come from.
+   * @return    A new table name map with unknown tables filtered out
+   *            (unless only unknown tables are present, in which case it
+   *            leaves an arbitrary unknown table which is assumed to contain
+   *            the column we are trying to resolve).
+   *            This ensures that columns are not matched to unknown tables if
+   *            they can instead be matched to tables with known schemas.
+   */
+  public Map<String, ScopeChild> filterTableNames(Map<String, ScopeChild> map) {
+    Map<String, ScopeChild> filteredMap = new HashMap<String, ScopeChild>();
+    for (String key : map.keySet()) {
+      if (!containsUnknownNamespace(map.get(key))) {
+        filteredMap.put(key, map.get(key));
+      }
+    }
+    // If original map consists only of unknown tables,
+    // arbitrarily assume the column corresponds to the first unknown table.
+    // We cannot do any better than this without prior schema information.
+    if (filteredMap.size() == 0 && map.size() > 0) {
+      String arbitraryKey = map.keySet().iterator().next();
+      filteredMap.put(arbitraryKey, map.get(arbitraryKey));
+    }
+    return filteredMap;
+  }
+
+  /**
+   * Determines whether the given {@code ScopeChild} corresponds to the
+   * namespace of an unknown table.
+   */
+  public boolean containsUnknownNamespace(ScopeChild child) {
+    if (child.namespace != null) {
+      SqlValidatorNamespace resolvedNamespace = child.namespace.resolve();
+      return resolvedNamespace.getType() instanceof UnknownRecordType;
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -214,6 +216,47 @@ public SqlValidator getValidator() {
     return validator;
   }
 
+  /**
+   *
+   * @param map A map from table names to {@code ScopeChild} objects, which
```
dasch-1, line 221: Let's add a top-level sentence to this function's javadoc.

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -214,6 +216,47 @@ public SqlValidator getValidator() {
...
+   *            they can instead be matched to tables with known schemas.
+   */
+  public Map<String, ScopeChild> filterTableNames(Map<String, ScopeChild> map) {
+    Map<String, ScopeChild> filteredMap = new HashMap<String, ScopeChild>();
+    for (String key : map.keySet()) {
+      if (!containsUnknownNamespace(map.get(key))) {
+        filteredMap.put(key, map.get(key));
+      }
+    }
+    // If original map consists only of unknown tables,
```
dasch-1, line 238: nit: This line looks like it's broken before the 80 character cut-off and could have had the word "arbitrarily" without going over. Please adjust this comment (and others throughout the PR) such that a line contains as many words as possible (without going over 80 characters) before going to the next line.

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -214,6 +216,47 @@ public SqlValidator getValidator() {
...
+    // arbitrarily assume the column corresponds to the first unknown table.
+    // We cannot do any better than this without prior schema information.
+    if (filteredMap.size() == 0 && map.size() > 0) {
```
dasch-1, line 241: nit: Use isEmpty() functions (or equivalent) for both of these checks.

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -214,6 +216,47 @@ public SqlValidator getValidator() {
...
+    }
+    return filteredMap;
+  }
+
+  /**
+   * Determines whether the given {@code ScopeChild} corresponds to the
+   * namespace of an unknown table.
+   */
+  public boolean containsUnknownNamespace(ScopeChild child) {
+    if (child.namespace != null) {
```
dasch-1, line 253: nit: Negate this if-check and return early. Then you can unnest the logic within it.

core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -170,16 +177,38 @@ private SqlValidatorNamespace resolveImpl(SqlIdentifier id) {
...
       }
     }
+    if (validator.config().handleUnknownTables()) {
+      return getPlaceholderNamespace(id);
+    }
     throw validator.newValidationError(id,
         RESOURCE.objectNotFound(id.getComponent(0).toString()));
   }
 
+  /**
```
dasch-1, line 196: nit: Add top-level javadoc sentence for this function.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
```diff
@@ -931,6 +931,18 @@ boolean validateModality(SqlSelect select, SqlModality modality,
      */
     Config withCallRewrite(boolean rewrite);
 
+    /**
+     * Returns whether to allow unresolved types during validation.
+     */
+    @ImmutableBeans.Property
+    @ImmutableBeans.BooleanDefault(false)
+    boolean handleUnknownTables();
```
dasch-1, line 939: nit: /handleUnknownTables/allowUnknownTables/ ?

core/src/test/java/org/apache/calcite/sql/test/SqlTestFactory.java
```diff
@@ -133,11 +133,20 @@ public SqlValidator getValidator() {
         (SqlConformance) options.get("conformance");
     final boolean lenientOperatorLookup =
         (boolean) options.get("lenientOperatorLookup");
-    final boolean enableTypeCoercion = (boolean) options.get("enableTypeCoercion");
+    final boolean enableTypeCoercion =
+        (boolean) options.get("enableTypeCoercion");
+    final boolean handleUnknownTables;
+    if (options.containsKey("handleUnknownTables")) {
```
dasch-1, line 139: The other options don't seen to check for existence before calling get(). Can we just call get() directly? Will it default to returning false if the key is not found?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
dstekol: Sure, I can change it to allowUnknownTables

core/src/test/java/org/apache/calcite/sql/test/SqlTestFactory.java
dstekol: It will fail if we call get() on a nonexistent key. I believe the other options are not checked beforehand because they are always set explicitly. I figured it would be easiest to have handleUnknownTables be optional and just check that it is in the map before retrieving it, but I could change the existing code to always set the handleUnknownTables option explicitly to either true or false if you think that is preferable.

core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -170,16 +177,37 @@ private SqlValidatorNamespace resolveImpl(SqlIdentifier id) {
...
       }
     }
+    if (validator.config().allowUnknownTables()) {
+      return getPlaceholderNamespace(id);
+    }
     throw validator.newValidationError(id,
         RESOURCE.objectNotFound(id.getComponent(0).toString()));
   }
 
+  /** Returns a "placeholder" namespace containing a table with an
```
AndrewPochapsky, line 196: nit: Indentation looks off, should this be on a new line?

core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -170,16 +177,37 @@ private SqlValidatorNamespace resolveImpl(SqlIdentifier id) {
...
+   * {@code UnknownRecordType}. This allows us to treat unknown table
+   * identifiers as if they are dynamic tables when the {@allowUnknownTables}
+   * option of the validator's config object is enabled.
+   * @param id  An identifer which cannot be resolved.
```
AndrewPochapsky, line 200: nit: Add newline before @param

core/src/test/java/org/apache/calcite/sql/test/SqlTestFactory.java
dasch-1: To keep things consistent, let's follow the same pattern as is used for all of the other options.

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -214,6 +216,47 @@ public SqlValidator getValidator() {
...
+  /**
+   * Determines whether the given {@code ScopeChild} corresponds to the
+   * namespace of an unknown table.
+   */
+  public boolean containsUnknownNamespace(ScopeChild child) {
+    if (child.namespace == null) {
+      return false;
+    }
+    SqlValidatorNamespace resolvedNamespace = child.namespace.resolve();
+    return resolvedNamespace.getType() instanceof UnknownRecordType;
```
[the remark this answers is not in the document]
dstekol: I haven't tried this but I don't think it would: resolvedNamespace is an instance of SqlValidatorNamespace, and getType() returns an instance of RelDataType. I can understand where you're coming from, but getType() isn't a built-in method returning the type of resolvedNamespace, it's actually a class method returning the row type that the object represented by the namespace corresponds to - so if resolvedNamespace corresponds to the namespace of table FOO, which has one VARCHAR column, getType() will return something like RecordType(VARCHAR(20)).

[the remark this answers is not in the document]
AndrewPochapsky: Ah ok, thanks for the clarification. I thought that getType() was built in.

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -241,6 +241,7 @@ public SqlValidator getValidator() {
     if (filteredMap.isEmpty() && !map.isEmpty()) {
       String arbitraryKey = map.keySet().iterator().next();
       filteredMap.put(arbitraryKey, map.get(arbitraryKey));
+      //todo(dstekol): Figure out how to represent ambiguous columns
```
dasch-1, line 244: nit: "// TODO(dstekol):"

PR #222, opened 2020-07-28
[CALCITE] Added support for caret negation syntax

michaelvdow:
Added support for "^" operator


parsing/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -2780,6 +2780,21 @@ SqlNode TableRefOrJoinClause() :
...
 
+void CaretNegation(ExprContext exprContext, List<Object> list) :
+{
+    final SqlOperator op;
+}
+{
+    <CARET>
+    Expression2b(exprContext, list)
+    <EQ> {
+        op = SqlStdOperatorTable.NOT_EQUALS;
```
dasch-1, line 2791: nit: Can we just omit the temporary variable altogether and pass SqlStdOperatorTable.NOT_EQUALS directly into the ToTreeListItem() call?

michaelvdow: Yeah I'll update that

parsing/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -3662,4 +3662,28 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
         "CREATE TABLE `FOO` (`BAR` TIMESTAMP(6) WITH TIME ZONE)";
     sql(sql).ok(expected);
   }
+
+  @Test public void testCaretNegation() {
+    String sql = "select * from foo where ^a = 1";
```
lisayin, line 3667: does the caret operator work on expressions? eg ^(col == "something)

michaelvdow: I'm not sure since we couldn't find any docs for this. @dasch-1 Do you know if this would be valid?

dasch-1: Double equals is not valid in Dialect1. However, the caret operator does work with parens. More information can be found in this comment thread:

https://docs.google.com/document/d/1L0MK9dfWril2ppxJjCeCTL5p811Pl1oLLsxFCQdxhVI/edit?disco=AAAAGwhXSXg

michaelvdow: This is updated now so it uses a SqlCaretNegationOperator to unparse the "^" operator

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.calcite.sql.SqlWriter;
+
+/**
+ * SqlCaretNegationOperator represents the "^" operator that occurs before a
+ * binary row expression.
+ */
+public class SqlCaretNegationOperator extends SqlPrefixOperator {
+
+  public SqlCaretNegationOperator() {
+    super("^", SqlKind.CARET_NEGATION, 26, null, null, null);
```
dasch-1, line 31: nit: `/*prec=*/26`

parsing/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -2780,6 +2780,19 @@ SqlNode TableRefOrJoinClause() :
...
 }
 
+/**
+ * Parses a prefix row operator like NOT.
+ */
+SqlPrefixOperator PrefixRowOperator() :
+{}
+{
+    <PLUS> { return SqlStdOperatorTable.UNARY_PLUS; }
+|   <MINUS> { return SqlStdOperatorTable.UNARY_MINUS; }
```
dasch-1, line 2790: nit: I know this matches the style in the original PrefixRowOperator, but I'd suggest making this OR style match what's predominantly used elsewhere in this file (e.g. what's used in TableRefOrJoinClause() above).

michaelvdow: Sure that sounds good, will update

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
dstekol: It would probably be good to add comments for the other params too.

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override public void unparse(SqlWriter writer, SqlCall call, int leftPrec,
+      int rightPrec) {
+    writer.keyword(getName());
+    writer.setNeedWhitespace(false);
+    call.operand(0).unparse(writer, getLeftPrec(), getRightPrec());
```
dstekol, line 38: What is the difference between calling getLeftPrec() and just passing in the leftPrec parameter which was passed to this method?

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.calcite.sql.SqlWriter;
+
+/**
+ * SqlCaretNegationOperator represents the "^" operator that occurs before a
+ * binary row expression.
+ */
+public class SqlCaretNegationOperator extends SqlPrefixOperator {
+
+  public SqlCaretNegationOperator() {
+    super("^", SqlKind.CARET_NEGATION, /*prec=*/26, null, null, null);
```
dstekol, line 31: Maybe it would also be good to add inline comments for the nulls here?

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
dstekol: It would probably be good to also add inline comments for the other parameters here.

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
michaelvdow: I think I added those on accident, it works with leftPrec/rightPrec so I'll switch those calls to variables since that is what's usually used

core/src/main/java/org/apache/calcite/sql/fun/SqlCaretNegationOperator.java
michaelvdow: Sure, I updated them with comments

dasch-1:
Added support for "^" operator.

PR #223, opened 2020-07-28
[CALCITE] First part of Keywords Refactor

AndrewPochapsky:
There are a lot of changes so I decided to split it up into multiple PRs. This one contains:
- The processKeywords() method + tests 
- The new Keyword class + tests
- The unparseReservedKeywords() method + tests

### DIFF
--- a/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
+++ b/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
@@ -21,6 +21,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.Set;
 import java.util.regex.MatchResult;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -55,6 +56,62 @@ public class DialectGenerate {
 
   public static Queue<String> getTokens(String input) {
     return new LinkedList<String>(Arrays.asList(TOKENIZER_PATTERN.split(input)));
+  }
+
+  /**
+   * Adds specified keywords and nonReservedKeywords to extractedData. Also
+   * ensures that nonReservedKeywords is a subset of the union of keywords
+   * and extractedData.keywords.
+   *
+   * @param keywords The keywords to add
+   * @param nonReservedKeywords the non reserved keywords to add
+   * @param extractedData The object to which the keywords will be added to
+   */
+  public void processKeywords(Map<String, String> keywords,
+      Set<String> nonReservedKeywords,
+      ExtractedData extractedData) throws IllegalStateException {
+    for (String keyword : nonReservedKeywords) {
+      if (!(keywords.containsKey(keyword)
+            || extractedData.keywords.containsKey(keyword))) {
+        throw new IllegalStateException(keyword + " is not a keyword.");
+      }
+    }
+    extractedData.keywords.putAll(keywords);
+    extractedData.nonReservedKeywords.addAll(nonReservedKeywords);
+  }
+
+  /**
+   * Adds extractedData.keywords (if nonempty) to extractedData.tokenAssignments with the
+   * form:
+   * <DEFAULT, DQID, BTID> TOKEN :
+   * {
+   *    <TOKEN_1: "TOKEN_1_VALUE">
+   *   |<TOKEN_2: "TOKEN_2_VALUE">
+   *   ...
+   * }
+   *
+   * @param extractedData The object which keeps state of all of the extracted
+   *                      data
+   */
+  public void unparseReservedKeywords(ExtractedData extractedData) {
+    if (extractedData.keywords.size() == 0) {
+      return;
+    }
+    boolean first = true;
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN : \n{\n");
+    for (Map.Entry<String, String> entry : extractedData.keywords.entrySet()) {
+      String token = String.format("<%s : \"%s\">\n", entry.getKey(),
+          entry.getValue());
+      if (!first) {
+        token = "|" + token;
+      } else {
+        first = false;
+      }
+      stringBuilder.append(token);
+    }
+    stringBuilder.append("}\n");
+    extractedData.tokenAssignments.add(stringBuilder.toString());
   }
 
   /**
--- a/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
+++ b/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
@@ -102,26 +102,24 @@ public class DialectGenerate {
     if (extractedData.keywords.isEmpty()) {
       return;
     }
-    boolean first = true;
     StringBuilder stringBuilder = new StringBuilder();
     stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN :\n{\n");
+    String tokenTemplate = "<%s : \"%s\">";
+    List<String> tokens = new LinkedList<String>();
     for (Map.Entry<Keyword, String> entry : extractedData.keywords.entrySet()) {
+      StringBuilder tokenBuilder = new StringBuilder();
       Keyword keyword = entry.getKey();
-      String tokenTemplate = "| <%s : \"%s\">";
-      if (first) {
-        tokenTemplate = "<%s : \"%s\">";
-        first = false;
-      }
-      String token = String.format(tokenTemplate, keyword.keyword,
-          entry.getValue());
-      stringBuilder.append(token);
+      tokenBuilder.append(String.format(tokenTemplate, keyword.keyword,
+          entry.getValue()));
       if (keyword.filePath == null) {
-        stringBuilder.append(" // No file specified.");
+        tokenBuilder.append(" // No file specified.");
       } else {
-        stringBuilder.append(" // From: ").append(keyword.filePath);
+        tokenBuilder.append(" // From: ").append(keyword.filePath);
       }
-      stringBuilder.append("\n");
+      tokenBuilder.append("\n");
+      tokens.add(tokenBuilder.toString());
     }
+    stringBuilder.append(String.join("| ", tokens));
     stringBuilder.append("}\n");
     extractedData.tokenAssignments.add(stringBuilder.toString());
   }
--- a/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
+++ b/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
@@ -66,6 +66,9 @@ public class DialectGenerate {
    * @param keywords The keywords to add
    * @param nonReservedKeywords the non reserved keywords to add
    * @param extractedData The object to which the keywords will be added to
+   *
+   * @throws IllegalStateException when an an element in nonReservedKeywords is
+   *                               not in keywords or extractedData.keywords
    */
   public void processKeywords(Map<Keyword, String> keywords,
       Set<Keyword> nonReservedKeywords,
--- a/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
+++ b/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
@@ -89,6 +89,8 @@ public class DialectGenerate {
    *   |<TOKEN_2: "TOKEN_2_VALUE">
    *   ...
    * }
+   * File annotations are added as single-line comments following each token
+   * if the filePath is specified.
    *
    * @param extractedData The object which keeps state of all of the extracted
    *                      data
@@ -99,9 +101,10 @@ public class DialectGenerate {
     }
     boolean first = true;
     StringBuilder stringBuilder = new StringBuilder();
-    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN : \n{\n");
+    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN :\n{\n");
     for (Map.Entry<Keyword, String> entry : extractedData.keywords.entrySet()) {
-      String token = String.format("<%s : \"%s\">\n", entry.getKey().keyword,
+      Keyword keyword = entry.getKey();
+      String token = String.format("<%s : \"%s\">", keyword.keyword,
           entry.getValue());
       if (!first) {
         token = "|" + token;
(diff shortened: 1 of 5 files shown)


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
     return new LinkedList<String>(Arrays.asList(TOKENIZER_PATTERN.split(input)));
   }
 
+  /**
+   * Adds specified keywords and nonReservedKeywords to extractedData. Also
+   * ensures that nonReservedKeywords is a subset of the union of keywords
+   * and extractedData.keywords.
+   *
+   * @param keywords The keywords to add
+   * @param nonReservedKeywords the non reserved keywords to add
```
dasch-1, line 67: nit: /the/The/

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+   * @param extractedData The object to which the keywords will be added to
+   *
+   * @throws IllegalStateException when an an element in nonReservedKeywords is
```
dasch-1, line 70: nit: /when/When/

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+   *                               not in keywords or extractedData.keywords
+   */
+  public void processKeywords(Map<Keyword, String> keywords,
+      Set<Keyword> nonReservedKeywords,
+      ExtractedData extractedData) throws IllegalStateException {
+    for (Keyword keyword : nonReservedKeywords) {
+      if (!(keywords.containsKey(keyword)
+            || extractedData.keywords.containsKey(keyword))) {
```
dasch-1, line 78: Why does the existence of "keyword" in extractedData.keywords result in a IllegalStateException(" is not a keyword.") exception being thrown?

Perhaps these should be split out into two separate exception/error messages?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+   *   ...
+   * }
+   * File annotations are added as single-line comments following each token
+   * if the filePath is specified.
+   *
+   * @param extractedData The object which keeps state of all of the extracted
+   *                      data
+   */
+  public void unparseReservedKeywords(ExtractedData extractedData) {
+    if (extractedData.keywords.size() == 0) {
```
dasch-1, line 102: nit: extractedData.keywords.isEmpty()

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/ExtractedData.java
```diff
@@ -20,15 +20,21 @@
 import java.util.Map;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
+import java.util.Set;
```
dasch-1, line 23: nit: Out of order alphabetically

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/Keyword.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Simple container class for a keyword.
+ */
+public class Keyword {
+  public final String keyword;
+  public final String filePath;
+
+  public Keyword(String keyword) {
+    this(keyword, null);
```
dasch-1, line 30: nit: `/*filepath=*/null`

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/Keyword.java
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param keyword The keyword string
+   * @param filePath The file where this keyword was taken from
+   */
+  public Keyword(String keyword, String filePath) {
+    this.keyword = Objects.requireNonNull(keyword.toUpperCase());
+    this.filePath = filePath;
+  }
+
+  @Override
```
dasch-1, line 44: nit: Put @Override on the same line as the function declaration.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/Keyword.java
```diff
@@ -0,0 +1,63 @@
...
+  public int hashCode() {
+    // The filePath should not be considered when calculating hashCode.
+    return keyword.hashCode();
+  }
+
+  @Override
```
dasch-1, line 50: nit: Put @Override on the same line as the function declaration.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -24,10 +24,13 @@
 import java.util.Arrays;
 import java.util.Map;
 import java.util.LinkedHashMap;
+import java.util.HashSet;
```
dasch-1, line 27: nit: Out of order alphabetically. Can you also fix the order of the other imports above?

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -255,4 +284,102 @@ private void assertFileProcessed(String testName) {
...
+  }
+
+  @Test public void testProcessKeywordsInvalidNonReservedKeywordFails() {
+    Map<Keyword, String> keywords = new LinkedHashMap<Keyword, String>();
+    ExtractedData extractedData = new ExtractedData();
+    Set<Keyword> nonReservedKeywords = new HashSet<Keyword>();
+
+    keywords.put(new Keyword("foo"), "foo");
+    nonReservedKeywords.add(new Keyword("bar"));
+    assertKeywordsNotProcessed(keywords, nonReservedKeywords, extractedData);
```
dasch-1, line 342: nit: Can we just pass "new ExtractedData()" directly in here as is being done in testProcessKeywordsBothNonEmptyAndValid()?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+    }
+    boolean first = true;
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN :\n{\n");
+    for (Map.Entry<Keyword, String> entry : extractedData.keywords.entrySet()) {
+      Keyword keyword = entry.getKey();
+      String token = String.format("<%s : \"%s\">", keyword.keyword,
+          entry.getValue());
+      if (!first) {
+        token = "|" + token;
```
michaelvdow, line 113: I think StringBuilder be used here as well for the "|" if it is checked before token is added

dasch-1, line 113: nit: Let's add a space after the "|" symbol (e.g. "| ")

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/KeywordTest.java
```diff
@@ -0,0 +1,64 @@
...
+public class KeywordTest {
+
+  @Test public void testHashCodeEquality() {
+    Keyword a = new Keyword("foo");
+    Keyword b = new Keyword("foo", "path/file");
+    assertEquals(a.hashCode(), b.hashCode());
+    assertEquals(a, a);
+  }
+
+  @Test public void testHashCodeInEquality() {
```
dasch-1, line 35: nit: testHashCodeInequality (since "inequality" is one word).

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/KeywordTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class KeywordTest {
+
+  @Test public void testHashCodeEquality() {
+    Keyword a = new Keyword("foo");
+    Keyword b = new Keyword("foo", "path/file");
+    assertEquals(a.hashCode(), b.hashCode());
+    assertEquals(a, a);
```
dasch-1, line 32: Why is this assertion here? It doesn't seem to go along with the name of the test.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/KeywordTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test public void testEquality() {
+    Keyword a = new Keyword("foo");
+    Keyword b = new Keyword("foo", "path/file");
+    Keyword c = new Keyword("FOO");
+    assertEquals(a, b);
+    assertEquals(a, c);
+  }
+
+  @Test public void testInEquality() {
```
dasch-1, line 49: nit: testInequality

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/KeywordTest.java
```diff
@@ -0,0 +1,64 @@
...
+    Keyword a = new Keyword("foo");
+    Keyword b = new Keyword("bar");
+    assertNotEquals(a, b);
+    assertNotEquals(a, null);
+  }
+
+  @Test public void testKeywodGetsCapitalized() {
```
dasch-1, line 56: nit: /testKeywodGetsCapitalized/testKeywordGetsCapitalized/

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/KeywordTest.java
AndrewPochapsky: Good catch, I meant to put that in a different test.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: I think you may have missed the negation there, the behaviour is that if a nonReservedKeyword is not in keywords and not in extractedData.keywords an exception should be thrown.

dasch-1: You're right, I did miss the parenthesized negation :)

nit: In that case, it might be a bit more readable without the extra parentheses. For example:

if (!keywords.containsKey(keyword) && !extractedData.keywords.containsKey(keyword))

AndrewPochapsky: Sounds good, will make the change.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+      return;
+    }
+    boolean first = true;
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN :\n{\n");
+    for (Map.Entry<Keyword, String> entry : extractedData.keywords.entrySet()) {
+      Keyword keyword = entry.getKey();
+      String tokenTemplate = "| <%s : \"%s\">";
+      if (first) {
+        tokenTemplate = "<%s : \"%s\">";
```
lisayin, line 112: Does the first token have an ending >? In the example comment it doesnt.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -57,6 +58,74 @@
...
+  public void unparseReservedKeywords(ExtractedData extractedData) {
+    if (extractedData.keywords.isEmpty()) {
+      return;
+    }
+    boolean first = true;
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append("<DEFAULT, DQID, BTID> TOKEN :\n{\n");
+    for (Map.Entry<Keyword, String> entry : extractedData.keywords.entrySet()) {
+      Keyword keyword = entry.getKey();
+      String tokenTemplate = "| <%s : \"%s\">";
```
lisayin, line 110: Can you use a string join function after the loop that uses '|' as a delimiter to avoid having to special case the first token?

AndrewPochapsky: Good idea, changed logic to use String.join.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -102,26 +102,24 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+        tokenBuilder.append(" // No file specified.");
       } else {
-        stringBuilder.append(" // From: ").append(keyword.filePath);
+        tokenBuilder.append(" // From: ").append(keyword.filePath);
       }
-      stringBuilder.append("\n");
+      tokenBuilder.append("\n");
+      tokens.add(tokenBuilder.toString());
     }
+    stringBuilder.append(String.join("| ", tokens));
```
dasch-1, line 122: Does the current StringJoin logic work with the comments above? I may just be reading this incorrectly, but it seems like it would put the "|" character after the comment?

dasch-1: Maybe we can include a test in this PR to illustrate that this works?

AndrewPochapsky: It does put it after but there is a new line after each token so it works out.

dasch-1: Ah true.

AndrewPochapsky: There is already a test here: https://github.com/googleinterns/calcite/blob/619cd86e3738c1a74de61b176947b2655544692f/buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java#L367

dasch-1: Thanks, I missed that too :)

PR #230, opened 2020-07-29
[CALCITE] Add support for TRANSLATE_CHK function

dasch-1:


dasch-1:
The TRANSLATE_CHK built-in function determines if a TRANSLATE conversion can be performed without producing errors. It returns an integer test result. This can be used to filter untranslatable strings.

core/src/main/java/org/apache/calcite/sql/SqlTranslateUsingCharacterSet.java
```diff
@@ -27,29 +28,29 @@
...
    * @param pos  Parser position, must not be null
    * @param args  List of SqlNode, the first one is an char sequence to be
    *              translated, the second one is the AST node of SqlCharacterSetToCharacterSet
-   * @param isWithError If the with error token is specified
+   * @param isTranslateChk  If the TRANSLATE_CHK function should be used instead of TRANSLATE
+   * @param isWithError  If the WITH ERROR tokens are specified
    */
-  public SqlTranslateUsingCharacterSet(final SqlParserPos pos, final List<SqlNode> args,
-      final boolean isWithError) {
+  public SqlTranslateUsingCharacterSet(SqlParserPos pos, List<SqlNode> args, boolean isTranslateChk,
```
lisayin, line 44: nit: line seems long

PR #232, opened 2020-07-29
[CALCITE] Added support for ALTER PROCEDURE statements

michaelvdow:


parsing/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4079,6 +4079,8 @@ SqlAlter SqlAlter() :
         alterNode = SqlAlterTable(s, scope)
     |
         alterNode = SqlSetOption(s, scope)
+    |
+        alterNode = SqlAlterProcedure(s, scope)
```
dasch-1, line 4083: nit: Can we put these clauses in alphabetical order?

michaelvdow: Updated

core/src/main/java/org/apache/calcite/sql/SqlAlterProcedure.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Override protected void unparseAlterOperation(SqlWriter writer, int leftPrec,
+      int rightPrec) {
+    procedureName.unparse(writer, 0, 0);
+    if (languageSql) {
+      writer.keyword("LANGUAGE SQL");
+    }
+    writer.keyword("COMPILE");
+    if (!options.isEmpty()) {
```
dstekol, line 74: This could throw a NullPointerException if options is null - might be worth adding an Objects.requireNonNull in constructor for this. Also applies to procedureName.

michaelvdow: Yeah that's a good point, I updated them with requireNonNull calls

PR #234, opened 2020-07-29
[CALCITE] Second part of Keywords Refactor

AndrewPochapsky:
Builds on top of https://github.com/googleinterns/calcite/pull/223
- Added unparsing logic for non reserved keywords
- Added annotation "// Auto generated" above each function/token assignment created as a part of this feature
- Modified existing tests to work with annotation, added new tests for unparsing logic


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +125,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
       tokens.add(tokenBuilder.toString());
     }
     stringBuilder.append(String.join("| ", tokens));
-    stringBuilder.append("}\n");
+    stringBuilder.append("}");
```
ArthurInTheShell, line 128: This is more of an open question. Is there any side effect introduced in the change here to not add the next line character in the resulted generated file?

AndrewPochapsky: Good question. I made this change to stay consistent with how the tokens/functions are unparsed in DialectTraverser here: https://github.com/googleinterns/calcite/blob/733e88cb24142b26dd8e89d5e7bd9fbe0bb793b3/buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java#L94

As you can see the newline is being added there so adding it here would result in a double new line.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+
+  /**
+   * Partitions {@code keywords} into sets of size at most 500.
+   *
+   * @param keywords The keywords to partition
+   *
+   * @return The partitioned keywords
+   */
+  private ArrayList<Set<Keyword>> getPartitions(Set<Keyword> keywords) {
+    final int partitionSize = 500;
```
dasch-1, line 233: nit: Let's define the partition size as a private static final in the class.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+  }
+
+  /**
+   * Partitions {@code keywords} into sets of size at most 500.
+   *
+   * @param keywords The keywords to partition
+   *
+   * @return The partitioned keywords
+   */
+  private ArrayList<Set<Keyword>> getPartitions(Set<Keyword> keywords) {
```
dasch-1, line 232: nit: Should we mark "keywords" as final to make it clear that we don't plan to modify it in this function?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+    final int partitionSize = 500;
+    final int numPartitions = (int) Math.ceil(((double) keywords.size()) / partitionSize);
+    final ArrayList<Set<Keyword>> partitions = new ArrayList<Set<Keyword>>();
```
dasch-1, line 235: nit: /`new ArrayList<Set<Keyword>>`/`new ArrayList<>`/

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+    final int numPartitions = partitions.size();
+    final String declarationTemplate = "void NonReservedKeyword%dof"
+      + numPartitions + "():";
+    final StringBuilder bodyBuilder = new StringBuilder();
+    final List<String> functionCalls = new LinkedList<String>();
+    if (!partitions.isEmpty()) {
+      bodyBuilder.append("(\n");
+    }
+    for (int i = 0; i < numPartitions; i++) {
+      String declaration = String.format(declarationTemplate, i, numPartitions);
```
dasch-1, line 168: What does "numPartitions" do in this line? It looks like "declarationTemplate" only has one format string (for "i").

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+      String functionName = getFunctionName(declaration);
```
dasch-1, line 169: This seems a little counterintuitive to use a regex to extract the function name when we already know what it is ("NonReservedKeyword%dof" % i). It looks like we're doing this already on line 182 for the top-level NonReservedKeyword function.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+   *
+   * Note: Indentation is added here just for clarity, actual function doesn't
+   * have indentation as it is difficult to format using String.join().
+   *
+   * @param declaration The function declaration
+   * @param keywords The keywords to be in the function body
+   *
+   * @return The formatted string
+   */
+  public String getPartitionFunction(String declaration, Set<Keyword> keywords) {
```
dasch-1, line 206: nit: Should we mark "keywords" as final?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+    final List<String> tokens = new LinkedList<String>();
```
dasch-1, line 207: nit: /`new LinkedList<String>`/`new LinkedList<>`/

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+    final StringBuilder bodyBuilder = new StringBuilder();
+    for (Keyword keyword : keywords) {
+      StringBuilder tokenBuilder = new StringBuilder();
+      tokenBuilder.append("<").append(keyword.keyword)
+        .append(">");
```
dasch-1, line 212: nit: Move this up to the line above.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/ExtractedData.java
```diff
@@ -17,8 +17,9 @@
 package org.apache.calcite.buildtools.parser;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashMap;
```
dasch-1, line 21: This is a duplicate import.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -379,13 +381,103 @@ private void assertFileProcessed(String testName, boolean specifyFilePath) {
...
+         + "}";
+    assertEquals(expected, actual);
+  }
+
+  @Test public void testPartitionFunctionMultiple() {
+    Set<Keyword> keywords = new LinkedHashSet<Keyword>();
+    DialectGenerate dialectGenerate = new DialectGenerate();
+    keywords.add(new Keyword("foo", "file/path"));
+    keywords.add(new Keyword("bar", /*filePath=*/ null));
+    keywords.add(new Keyword("baz", "file/path"));
```
dasch-1, line 413: nit: Use a different filepath for either foo or baz?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -120,10 +124,152 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
...
+   * {
+   *     (
+   *         NonReservedKeyword0of<numPartitions>()
+   *     |   NonReservedKeyword1of<numPartitions>()
+   *     ...
+   *     )
+   *     { return unquotedIdentifier(); }
+   * }
+   *
+   * If {@code extractedData.nonReservedKeywords} is empty, the function only contains
```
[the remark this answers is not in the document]
AndrewPochapsky: I suppose it would, I did this to ensure that this function always exists regardless of if nonReservedKeywords are specified. This is because it is referenced somewhere in parsing/parserImpls.ftl so not having it would prevent compiling. I'm not particularity attached to this solution, do you have a better approach?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Good catch, I forgot to remove that.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Good point, will change.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
AndrewPochapsky: Good idea.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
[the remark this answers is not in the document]
dasch-1: If there are no keywords, then you might want to just insert the <EOF> token here. So that would look like:

`void NonReservedKeyword():
{
}
{
  <EOF>
  { return unquotedIdentifier(); }
}
`

We should also add a test for this scenario of no nonreserved keywords.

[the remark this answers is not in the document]
AndrewPochapsky: Will do, there is already a test but I'll modify it to include this.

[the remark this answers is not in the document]
dasch-1: I meant a test that actually exercised the parsing codepath where we have no nonreserved keywords. For example, in a *ParserTest.java file for a dialect with no non-reserved keywords.

This might be overkill and unnecessary, but just wanted to clarify what was meant.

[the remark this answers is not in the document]
AndrewPochapsky: Ah ok I see what you mean. I feel like the existing unit test covers this scenario quite well but I can also create a new structure just to test this if you want.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectGenerate.java
```diff
@@ -148,8 +148,8 @@ public void unparseReservedKeywords(ExtractedData extractedData) {
    *     { return unquotedIdentifier(); }
    * }
    *
-   * If {@code extractedData.nonReservedKeywords} is empty, the function only contains
-   * the return statement.
+   * If {@code extractedData.nonReservedKeywords} is empty, the function
+   * only only checks for the <EOF> token.
```
dasch-1, line 152: nit: /only only/only/

PR #240, opened 2020-07-30
[CALCITE] Added support for DROP PROCEDURE statements

michaelvdow:


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3397,6 +3397,8 @@ SqlDrop SqlDrop() :
         drop = SqlDropView(s, replace)
     |
         drop = SqlDropFunction(s, replace)
+    |
+        drop = SqlDropProcedure(s)
```
dasch-1, line 3401: nit: While you're here, can you alphabetize all of these SqlDrop clauses?

michaelvdow: Sure, it's updated now

PR #241, opened 2020-07-30
[CALCITE] Final part of Keywords Refactor

AndrewPochapsky:
Builds on top of https://github.com/googleinterns/calcite/pull/234

Note: YAML was used here instead of JSON due to YAML supporting comments which we need for the license. 

- Added JSON parsing logic
- Modified parserTest directory structure to have keywords.yaml files
- Slightly changed the formatting of generated file
- Moved keywords and non reserved keywords into keywords.json files
- Moved default keywords out of parsing/parserImpls.ftl and removed injection logic for non reserved keywords
- Moved general token declarations from parsing/parserImpls.ftl back to Parser.jj due to order of declaration mattering for certain tests


buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -444,11 +444,11 @@ private void assertFileProcessed(String testName, boolean specifyFilePath) {
     extractedData.nonReservedKeywords.add(new Keyword("foo"));
     dialectGenerate.unparseNonReservedKeywords(extractedData);
     assertEquals(2, extractedData.functions.size());
-    assertTrue(extractedData.functions.containsKey("NonReservedKeyword"));
+    assertTrue(extractedData.functions.containsKey("NonReservedKeyWord"));
     assertTrue(extractedData.functions.containsKey("NonReservedKeyword0of1"));
-    String actual = extractedData.functions.get("NonReservedKeyword");
+    String actual = extractedData.functions.get("NonReservedKeyWord");
```
michaelvdow, line 449: nit: Could this be kept as "Keyword" since some variable names and autogenerated methods still have that format? Or the other way around to keep it constant

AndrewPochapsky: I'll try changing it back, I had some build issues before for some reason.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -141,21 +156,47 @@ private void traverse(
     String nextDirectory = directories.peek();
     for (File f : files) {
       String fileName = f.getName();
-      if (f.isFile() && fileName.endsWith(".ftl")) {
+      if (f.isFile()) {
```
dasch-1, line 159: Can we reverse the polarity of this if-check and continue early? That would allow us to unnest all of this logic, which appears to be larger than that in the else-if.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -141,21 +156,47 @@ private void traverse(
...
-          e.printStackTrace();
+        if (fileName.endsWith(".ftl")) {
+          try {
+            dialectGenerate.processFile(fileText, extractedData, filePath);
+          } catch (IllegalStateException e) {
+            e.printStackTrace();
+          }
+        } else if (fileName.equals("keywords.yaml")) {
+          // This is required to get past the license comment. The part after
+          // follows a JSON format is it can be parsed as JSON.
```
dasch-1, line 181: nit: This sentence reads a little strangely.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
```diff
@@ -461,7 +461,7 @@ private void assertFileProcessed(String testName, boolean specifyFilePath) {
   @Test public void testUnparseNonReservedKeywordsMultiplePartitions() {
     ExtractedData extractedData = new ExtractedData();
     DialectGenerate dialectGenerate = new DialectGenerate();
-    for (int i = 0; i < 501; i++) {
+    for (int i = 0; i < 251; i++) {
```
dasch-1, line 464: nit: Can we reference the static field in DialectGenerate here? e.g. through a public getter.

buildSrc/subprojects/parser/src/test/java/org/apache/calcite/test/DialectGenerateTest.java
AndrewPochapsky: I ended up changing them to all have the capital W, it was more straightforward.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
AndrewPochapsky: I'm not sure if that is possible given that there is an else-if after it.

dasch-1: Why can't we do the following?

`
if (!directories.isEmpty() && fileName.equals(nextDirectory)) {
  ...
  continue;
}
<code that was previously in if-clause>
`

AndrewPochapsky: This seems to be breaking when the traverser encounters a directory that isn't the `nextDirectory`.

dasch-1: I see. Then you should be able to fix that by updating the if-clause to:

`if (!f.isFile() && !directories.isEmpty() && fileName.equals(nextDirectory)) { ... continue; }`

This is just an optional suggestion. Just a thought that may make it a bit easier to read this by reducing the amount of nesting.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -141,21 +156,47 @@ private void traverse(
...
+          }
+        } else if (fileName.equals("keywords.yaml")) {
+          // This is required to get past the license comment. The part after
+          // follows a JSON format so it can be parsed as such.
+          fileText = fileText.substring(fileText.indexOf("{"));
+          JSONObject json = (JSONObject) new JSONTokener(fileText).nextValue();
+          JSONArray keywordsJson = json.isNull("keywords")
+            ? null
+            : json.getJSONArray("keywords");
+          JSONArray nonReservedKeywordsJson = json.isNull("nonReservedKeywords")
```
lisayin, line 187: some inconsistencies with using Keywords vs KeyWords in this change eg in DialectGenerate.java it uses KeyWords but here it's Keywords

dasch-1: My 2 cents: this should be ("keywords"/"Keywords") rather than ("keyWords"/"KeyWords") everywhere. This aligns with the pattern of only uppercasing letters that demarcate a new/distinct work in a variable name. However, "keyword" is just a single word.

AndrewPochapsky: I agree under normal circumstances but NonReservedKeyWords() (the function) is referenced deeply in Calcite and created some nasty build errors when I tried to change it. It also wasn't obvious where changes needed to be made.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
AndrewPochapsky: I also tried that and it fails for the same reason since that if statement can be false both when `f` is a directory and `f` is a file.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
dasch-1: Does it resolve the issues if you update the reference here (as well as the calls to the function in the parserImpls files)?

https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/parser/SqlAbstractParserImpl.java#L587

dasch-1: I'm not able to find any other references to this function name besides the one linked above, the one in the root parserImpls.ftl IdentifierSegment(), and one in Dialect1's SqlHostVariable().

Are there other references besides these?

AndrewPochapsky: Oh, that may work. I'll try it and report back.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
dasch-1: Hm, not sure I fully follow. But it doesn't really matter either way. Feel free to leave it as is for now.

PR #242, opened 2020-07-30
[CALCITE] Added support for basic BEGIN...END statements in CREATE PROCEDURE

michaelvdow:
Syntax added: `[<label_name>:] BEGIN <create_procedure_stmt_list> END [label_name]`
 - Created SqlStatementList to abstract unparsing logic for terminating semicolons
 - Created error message for when the labels don't match


core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
```diff
@@ -0,0 +1,49 @@
...
+ * statements with a terminating semicolon. It is also a {@link SqlNode}, so
+ * it may appear in a parse tree.
+ */
+public class SqlStatementList extends SqlNodeList {
+
+  public SqlStatementList(SqlParserPos pos) {
+    super(pos);
+  }
+
+  public SqlStatementList(Collection<? extends SqlNode> collection, SqlParserPos pos) {
```
dasch-1, line 34: nit: Line wrap

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4496,11 +4496,26 @@ SqlNode CreateProcedureStmt() :
...
     { return e; }
 }
 
+void CreateProcedureStmtList(SqlStatementList statements) :
+{
+    SqlNode e;
+}
+{
+    (
+        LOOKAHEAD(1)
```
dasch-1, line 4512: Why is LOOKAHEAD(1) required here?

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4582,3 +4597,19 @@ AlterProcedureWithOption AlterProcedureWithOption() :
...
+SqlBeginEndCall SqlBeginEndCall() :
+{
+    SqlIdentifier beginLabel = null;
+    SqlIdentifier endLabel = null;
+    final SqlStatementList statements = new SqlStatementList(getPos());
+    final Span s;
+}
+{
+    [ beginLabel = SimpleIdentifier() <COLON> ]
+    <BEGIN> { s = span(); }
```
dasch-1, line 4610: Why is the span() call here rather than above the beginLabel?

core/src/main/java/org/apache/calcite/sql/SqlBeginEndCall.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    if (label != null) {
+      label.unparse(writer, 0, 0);
+      writer.setNeedWhitespace(false);
+      writer.print(": ");
+    }
+    writer.keyword("BEGIN");
+    writer.newlineAndIndent();
```
dasch-1, line 63: How does newlineAndIndent() work? It looks like it indents by SqlPrettyWriter's currentIndent level. And currentIndent level is only modified when startList() is called. Is that correct?

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: When it was using the default lookahead of 2, it was throwing some errors saying it expected tokens like "INSERT ;" when I wouldn't expect it to lookahead for a semicolon. Changing it to LOOKAHEAD(CreateProcedureStmt()) also fixes it so it could be changed to that to make it more explicit?

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: It should probably be above, I'll update it

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
dasch-1: Ok, thanks for the explanation. Let's go with the LOOKAHEAD(CreateProcedureStmt()) option, since it's a bit more explicit.

core/src/main/java/org/apache/calcite/sql/SqlBeginEndCall.java
michaelvdow: Yeah it would indent when the startList is called above. In [SqlDialectParserTest](https://github.com/googleinterns/calcite/blob/master/parsing/src/test/java/org/apache/calcite/test/SqlDialectParserTest.java#L581) the indentation level is currently set to 0 so it doesn't indent and only creates a new line, but if that was set to 4 it would print with spaces correctly in the unit test

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: Yeah that sounds good, will update

core/src/main/java/org/apache/calcite/sql/SqlBeginEndCall.java
```diff
@@ -0,0 +1,78 @@
...
+   * @param beginLabel The label before the BEGIN keyword, may be null
+   * @param endLabel The label after the END keyword, may be null
+   * @param statements List of SQL statements
+   */
+  public SqlBeginEndCall(SqlParserPos pos, SqlIdentifier beginLabel,
+      SqlIdentifier endLabel, SqlStatementList statements) {
+    super(pos);
+    this.label = beginLabel;
+    this.statements = statements;
+    if (endLabel != null && !beginLabel.equalsDeep(endLabel, Litmus.IGNORE)) {
```
ArthurInTheShell, line 50: nit: Add null check to beginLabel. If we have a situation where the user only input endLabel but not beginLabel, then we might have a null pointer exception. I suggest we add a test for this case too.

michaelvdow: Good catch, I'll update that and add a test case

core/src/main/java/org/apache/calcite/sql/SqlBeginEndCall.java
```diff
@@ -0,0 +1,79 @@
...
+   * @param beginLabel The label before the BEGIN keyword, may be null
+   * @param endLabel The label after the END keyword, may be null
+   * @param statements List of SQL statements
+   */
+  public SqlBeginEndCall(SqlParserPos pos, SqlIdentifier beginLabel,
+      SqlIdentifier endLabel, SqlStatementList statements) {
+    super(pos);
+    this.label = beginLabel;
+    this.statements = statements;
+    if (endLabel != null && (beginLabel == null
```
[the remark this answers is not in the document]
michaelvdow: Yeah I can add a new line here

PR #243, opened 2020-07-31
[CALCITE] DELETE with delete_table_name

AndrewPochapsky:
Added support for queries of the form DELETE <delete_table_name> [FROM] table

### DIFF
--- a/parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
+++ b/parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
@@ -114,6 +114,18 @@ final class Dialect1ParserTest extends SqlDialectParserTest {
     sql(sql).ok(expected);
   }
 
+  @Test public void testDeleteWithTable() {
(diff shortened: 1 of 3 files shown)


parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -114,6 +114,37 @@
...
+  }
+
+  @Test public void testDeleteWithTableCompoundIdentifier() {
+    final String sql = "delete foo.bar from baz";
+    final String expected = "DELETE `FOO`.`BAR` FROM `BAZ`";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testDeleteWithTableWithAlias() {
+    final String sql = "delete foo from bar as b";
```
dstekol, line 130: Is it possible to have an alias for the first table as well?

AndrewPochapsky: I checked the docs and it isn't.

PR #244, opened 2020-07-31
[CALCITE] Added support for built in functions for inline expressions

michaelvdow:
Moved the inline expression calls out of NamedFunctionCall and placed them after AtomicRowExpression which parses both named and built in functions.


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3988,6 +3980,7 @@ SqlNode Expression3(ExprContext exprContext) :
     final SqlNodeList list2;
     final SqlOperator op;
     final Span s;
+    SqlNode inlineCall;
```
AndrewPochapsky, line 3983: nit: Could this be final?

michaelvdow: Yeah, it's updated to final now

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -4087,4 +4087,26 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
         + "TIME ZONE 'gmt'";
     sql(sql).ok(expected);
   }
+
+  @Test public void testAlternativeCastWithBuiltInFunction() {
+    final String sql = "select cast(a as date) (format 'yyy-mm-dd')";
```
dasch-1, line 4092: nit: /yyy/yyyy/

PR #245, opened 2020-07-31
[CALCITE] Script parsing: If Statement

ArthurInTheShell:
This feature help parses If Statements in control statements/stored procedures.


core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+/**
+ * Parse tree conditional expression paring with a list of statements .
```
dasch-1, line 24: nit: /paring/pairing/?

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public final SqlNodeList conditionalStmtListPairs;
+  public final SqlNodeList elseStmtList;
+
+  /**
+   * Creates a {@code SqlTablePartitionByColumn}.
+   * @param pos                       Parser position, must not be null.
+   * @param conditionalStmtListPairs  List of conditional expression pairs
+   *                                  with StatementList.
+   * @param elseStmtList              List of statement in the else clause.
```
dasch-1, line 38: nit: /statement/statements/

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
```diff
@@ -0,0 +1,73 @@
...
+   */
+  public SqlIfStmt(final SqlParserPos pos,
+      final SqlNodeList conditionalStmtListPairs,
+      final SqlNodeList elseStmtList) {
+    super(pos);
+    this.conditionalStmtListPairs = conditionalStmtListPairs;
```
dasch-1, line 44: nit: Should we assert that this is non-null?

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    for (int i = 0; i < conditionalStmtListPairs.size(); i++) {
+      if (i != 0) {
+        writer.keyword("ELSE IF");
+        conditionalStmtListPairs.get(i).unparse(writer, leftPrec, rightPrec);
+      } else {
+        writer.keyword("IF");
+        conditionalStmtListPairs.get(i).unparse(writer, leftPrec, rightPrec);
```
dasch-1, line 64: Since this is identical to line 61, let's move to after the if/else.

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public SqlStatementList(SqlParserPos pos) {
+    super(pos);
+  }
+
+  public SqlStatementList(Collection<? extends SqlNode> collection, SqlParserPos pos) {
+    super(collection, pos);
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec, final int rightPrec) {
```
dasch-1, line 38: nit: Line wrap

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
```diff
@@ -0,0 +1,49 @@
...
+ * statements with a terminating semicolon. It is also a {@link SqlNode}, so
+ * it may appear in a parse tree.
+ */
+public class SqlStatementList extends SqlNodeList {
+
+  public SqlStatementList(SqlParserPos pos) {
+    super(pos);
+  }
+
+  public SqlStatementList(Collection<? extends SqlNode> collection, SqlParserPos pos) {
```
dasch-1, line 34: nit: Line wrap

parsing/dialects/intermediate/dialects/dialect1/keywords.yaml
```diff
@@ -1,4 +1,4 @@
-# Licensed to the Apache Software Foundation (ASF) under one or more
+ # Licensed to the Apache Software Foundation (ASF) under one or more
```
dasch-1, line 1: nit: Remove extra space

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4582,3 +4584,65 @@ AlterProcedureWithOption AlterProcedureWithOption() :
...
+        return new SqlIfStmt(getPos(), conditionMultiStmtList,
+            elseMultiStmtList);
+    }
+}
+
+SqlNode ConditionMultiStmtPair() :
+{
+    final SqlNode condition;
+    final SqlStatementList multiStmtList = new SqlStatementList(getPos());
+    SqlNode e;
```
dasch-1, line 4625: nit: Remove unreferenced variable.

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
ArthurInTheShell: It cannot be null, base on the logic in the function IfStmt() in parserImpls.ftl. It would at least contain "IF <condition> THEN <statemment list>" part.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
dasch-1: That might be true for now. But isn't that making an assumption that the caller (e.g. the parser) never changes the way they construct this class? It's usually best practice to assert that it's non-null here.

parsing/dialects/intermediate/dialects/dialect1/keywords.yaml
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/01b6e805a3c5c8a60af83236b5e706211afdff7c

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
dasch-1: (This suggestion applies to all constructors added in this PR)

ArthurInTheShell: Sorry about the notification for this comment got ignored. Fixed in https://github.com/googleinterns/calcite/pull/245/commits/bc902df2d7f4882337c55eac53aa8dc3e7a31286.

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
```diff
@@ -41,6 +44,10 @@ public SqlIfStmt(final SqlParserPos pos,
       final SqlNodeList conditionalStmtListPairs,
       final SqlNodeList elseStmtList) {
     super(pos);
+    if (conditionalStmtListPairs == null) {
```
dasch-1, line 47: You should just be able to use Objects.requireNonNull() for this.

ArthurInTheShell: Sounds good. Added here https://github.com/googleinterns/calcite/pull/245/commits/0a11397894f8c800ebd81980b290f4a2dda11c7f

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+/**
+ * Parse tree conditional expression pairing with a list of statements .
```
dstekol, line 24: tiny nit: should probably be "Parse tree representing a conditional..."

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
```diff
@@ -0,0 +1,61 @@
...
+   * @param pos         Parser position, must not be null.
+   * @param condition   Condition expression.
+   * @param stmtList    A List of statements.
+   */
+  public SqlConditionalStmtListPair(final SqlParserPos pos,
+      final SqlNode condition,
+      final SqlStatementList stmtList) {
+    super(pos);
+    this.condition = condition;
+    this.stmtList = stmtList;
```
dstekol, line 44: Would be good to add requireNonNull() assertions here as well.

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
```diff
@@ -0,0 +1,51 @@
...
+      SqlParserPos pos) {
+    super(collection, pos);
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    SqlWriter.Frame frame = writer.startList(
+        SqlWriter.FrameTypeEnum.STATEMENT_LIST, "", "");
+    for (SqlNode e : getList()) {
+      e.unparse(writer, 0, 0);
```
dstekol, line 44: Might be good to add a null check for each list item as well before invoking a method on it, since even if the list itself is not null, the elements can be.

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
dstekol: nit: you can have the Objects.requireNonNull() call directly in the assignment like this (rather than on a separate line):
this.conditionalStmtListPairs = Objects.requireNonNull(conditionalStmtListPairs)
You could also do this for elseStmtList below.

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    for (int i = 0; i < conditionalStmtListPairs.size(); i++) {
+      if (i != 0) {
+        writer.keyword("ELSE IF");
+      } else {
+        writer.keyword("IF");
+      }
```
dstekol, line 66: You could potentially extract the "IF" keyword since it is present in both cases (so you would not need the else clause). Not a big deal, just a thought.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -4087,4 +4087,68 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
...
+        + "if abc = 3 then select baz; end if;"
+        + "end if";
+    final String expected =
+        "CREATE PROCEDURE `FOO` (IN `BEE` INTEGER, IN `ABC` INTEGER) "
+            + "IF (`BEE` = 2) THEN SELECT `BAR`;\n"
+            + "IF (`ABC` = 3) THEN SELECT `BAZ`;\n"
+            + "END IF;\n"
+            + "END IF";
+    sql(sql).ok(expected);
+  }
```
dstekol, line 4153: It would probably be useful to add a few failing test cases. For example, you cannot start an IF statement with an ELSE or an ELSE IF, and you cannot have an empty statement body (ex. "IF A>2 THEN END IF").

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
ArthurInTheShell: Thanks for the information on how to use Objects.requireNonNull(). I've checked Effective Java, and figured out I should start using it on all the Java classes methods' parameters that have non-null restriction.

ArthurInTheShell: For the "IF" keyword printing, I believe the existing logic looks more readable so I'll keep it. Unless some other reviewers have strong opinion on this.

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
ArthurInTheShell: Good catch. Fixed in https://github.com/googleinterns/calcite/pull/245/commits/a1a422eaffd994a532e1f41942f8295571c5d48b

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmtListPair.java
ArthurInTheShell: Fixed in https://github.com/googleinterns/calcite/pull/245/commits/a1a422eaffd994a532e1f41942f8295571c5d48b

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
ArthurInTheShell: It's a valid concern that the element of SqlStmtList could be null. After reading Effective Java item 49 for reference, I've decided to add non-null check for the add(SqlNode) method in SqlStatementList. The null check would then happen at storing instead of at consumption, which means it would fail early if it is meant to fail. The code change is in this commit https://github.com/googleinterns/calcite/pull/245/commits/a1a422eaffd994a532e1f41942f8295571c5d48b.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
ArthurInTheShell: Will add failing test soon.

ArthurInTheShell: Failing test cases added in https://github.com/googleinterns/calcite/pull/245/commits/ff9b2f8abfc814cbb8815c38bb8a55b35e5da747

core/src/main/java/org/apache/calcite/sql/SqlIfStmt.java
dasch-1: Leaving as is SGTM

dstekol: Fine either way, totally up to you.

core/src/main/java/org/apache/calcite/sql/SqlStatementList.java
dstekol: Even better.

PR #246, opened 2020-07-31
[CALCITE] Extended caret negation and NOT support before operators

michaelvdow:
- BinaryRowOperator overridden in Dialect1
 - Added support for caret negation operator "^" before` =, <, >, <=, >=, LIKE, IN`.
 - Added support for NOT before `=, <, >, <=, >=`
 - Unparses to equivalent existing operator

### DIFF
--- a/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
+++ b/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
@@ -3008,6 +3008,109 @@ SqlPrefixOperator PrefixRowOperator() :
 }
 
 /**
+ * Parses a binary row operator like AND.
+ */
+SqlBinaryOperator BinaryRowOperator() :
+{
+    SqlBinaryOperator op;
+}
+{
+    // <IN> is handled as a special case
+    (
+        <EQ> { return SqlStdOperatorTable.EQUALS; }
+    |
+        ( <GT> | (<CARET> | <NOT>) <LE> ) {
+            return SqlStdOperatorTable.GREATER_THAN;
+        }
+    |
+        ( <LT> | (<CARET> | <NOT>) <GE> ) {
+            return SqlStdOperatorTable.LESS_THAN;
+        }
+    |
+        ( <LE> | (<CARET> | <NOT>) <GT> ) {
+            return SqlStdOperatorTable.LESS_THAN_OR_EQUAL;
+        }
+    |
+        ( <GE> | (<CARET> | <NOT>) <LT> ) {
+            return SqlStdOperatorTable.GREATER_THAN_OR_EQUAL;
+        }
+    |
+        ( <NE> | (<CARET> | <NOT>) <EQ> )
+        {
+            return SqlStdOperatorTable.NOT_EQUALS;
+        }
+    |
+        <NE2> {
+            if (!this.conformance.isBangEqualAllowed()) {
+                throw SqlUtil.newContextException(getPos(),
+                    RESOURCE.bangEqualNotAllowed());
+            }
+            return SqlStdOperatorTable.NOT_EQUALS;
+        }
+    |
+        <PLUS> { return SqlStdOperatorTable.PLUS; }
+    |
+        <MINUS> { return SqlStdOperatorTable.MINUS; }
+    |
+        <STAR> { return SqlStdOperatorTable.MULTIPLY; }
+    |
+        <SLASH> { return SqlStdOperatorTable.DIVIDE; }
+    |
+        <PERCENT_REMAINDER> {
+            if (!this.conformance.isPercentRemainderAllowed()) {
+                throw SqlUtil.newContextException(getPos(),
+                    RESOURCE.percentRemainderNotAllowed());
+            }
+            return SqlStdOperatorTable.PERCENT_REMAINDER;
+        }
+    |
+        <CONCAT> { return SqlStdOperatorTable.CONCAT; }
+    |
+        <AND> { return SqlStdOperatorTable.AND; }
+    |
+        <OR> { return SqlStdOperatorTable.OR; }
+    |
+        LOOKAHEAD(2) <IS> <DISTINCT> <FROM> {
+            return SqlStdOperatorTable.IS_DISTINCT_FROM;
+        }
+    |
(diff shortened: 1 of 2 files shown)


parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -3922,6 +3922,102 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testCaretNegationEquals() {
+    String sql = "select * from foo where a ^= 1";
+    String expected = "SELECT *\n"
+        + "FROM `FOO`\n"
+        + "WHERE (`A` <> 1)";
```
dstekol, line 3929: Is it possible to negate not-equals (like ^<>)?

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -3922,6 +3922,102 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
...
+  @Test public void testNotGreaterThan() {
+    String sql = "select * from foo where a not > 1";
+    String expected = "SELECT *\n"
+        + "FROM `FOO`\n"
+        + "WHERE (`A` <= 1)";
+    sql(sql).ok(expected);
+  }
+
+  @Test public void testNotLessThanOrEqualTo() {
+    String sql = "select * from foo where a not <= 1";
```
dstekol, line 4006: Is it allowed to have both "not" and ^ together (not that I think this would ever be useful, but just wondering)? If it is not allowed, it might be good to add an extra test asserting that this fails

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3007,6 +3007,109 @@ SqlPrefixOperator PrefixRowOperator() :
...
 
+/**
+ * Parses a binary row operator like AND.
+ */
+SqlBinaryOperator BinaryRowOperator() :
+{
+    SqlBinaryOperator op;
+}
+{
+    // <IN> is handled as a special case
```
dasch-1, line 3018: Should we add to this comment that LIKE is also handled separately?

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
dasch-1: I don't see why not. Maybe we can add this on line 3020 in BinaryRowOperator()?

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
dasch-1: Good question. I've put out a request to a teammate for this information. Will reply here when I get the answer back.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
michaelvdow: Yeah I can add that in there along with a test case for it

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: Sure I'll update this

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
dasch-1: I just received confirmation that "^" and "NOT" cannot be used in conjunction (in any order) with these operators.

dstekol: Might still be good to add a failing test to clarify this, no?

dasch-1: SGTM

michaelvdow: Yeah I can add one in

michaelvdow: Actually I'm not sure how to escape the "^" since it uses those in a failing test case to indicate the token. Do you know how to escape it because the method (https://github.com/googleinterns/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java#L413) looks like it doesn't catch this case?

dstekol: Hmm, yeah, the testing framework might not have accounted for ^ being used in other cases. Maybe just put a //TODO for now?

dasch-1: Ah, I didn't even think of how the "^" is treated specially in the testing framework. I'd say either log a TODO or just do nothing at all for the proposed test case given the current shortcomings of the test environment for this specific syntax.

michaelvdow: Sure I could log a TODO here for when the test environment supports this.

PR #254, opened 2020-08-03
[CALCITE] Added support for local declarations in BEGIN...END clause

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlDeclareCondition.java
```diff
@@ -0,0 +1,66 @@
...
+   * Creates an instance of {@code SqlDeclareCondition}.
+   *
+   * @param pos SQL parser position
+   * @param conditionName Name of the declared condition
+   * @param stateCode SQLSTATE value assigned to condition, may be null
+   */
+  public SqlDeclareCondition(SqlParserPos pos, SqlIdentifier conditionName,
+      SqlNode stateCode) {
+    super(pos);
+    this.conditionName = conditionName;
```
dasch-1, line 45: nit: Objects.requireNotNull (here and throughout the rest of the PR)

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4622,16 +4622,84 @@ SqlBeginEndCall SqlBeginEndCall() :
...
+{
+    <DECLARE>
+    conditionName = SimpleIdentifier()
+    <CONDITION>
+    (
+        <FOR> stateCode = StringLiteral()
+    |
+        { stateCode = null; }
+    )
+    { return new SqlDeclareCondition(s.end(this), conditionName, stateCode); }
```
dasch-1, line 4700: nit: Line wrap

core/src/main/java/org/apache/calcite/sql/SqlDeclareCondition.java
michaelvdow: Forgot to add those in, it's updated now

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: It says this line is < 80 chars, did you want the return on a new line?

dasch-1: No, it looks like you're right. I must have miscounted.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4622,16 +4622,84 @@ SqlBeginEndCall SqlBeginEndCall() :
...
+    dataType = DataType()
+    (
+        <DEFAULT_>
+        (
+            defaultValue = Literal()
+        |
+            <NULL> { defaultValue = SqlLiteral.createNull(getPos()); }
+        )
+    |
+        { defaultValue = null; }
```
ArthurInTheShell, line 4677: This way of assigning null to a variable did eliminate the need for defaultValue to not be final, and it will make the assignment of null more explicit. I believe this new way of implementing it is the preferable way for implementing null-able variable. I'll probably adapt to this way of assigning it in my future PR. Thanks for figuring this out.

PR #255, opened 2020-08-03
[CALCITE] Added support for BEGIN REQUEST call in CREATE PROCEDURE

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlBeginRequestCall.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Creates an instance of {@code SqlBeginRequestCall}.
+   *
+   * @param pos SQL parser position
+   * @param statements List of SQL statements
+   */
+  public SqlBeginRequestCall(SqlParserPos pos, SqlStatementList statements) {
+    super(pos);
+    this.statements = statements;
```
dasch-1, line 44: nit: Objects.requireNotNull

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4695,3 +4697,19 @@ SqlRenameProcedure SqlRenameProcedure() :
...
+// Semicolon is optional after the last statement.
+SqlBeginRequestCall SqlBeginRequestCall() :
+{
+    final SqlStatementList statements = new SqlStatementList(getPos());
+    final Span s = Span.of();
+    SqlNode e;
+}
+{
+    <BEGIN> <REQUEST>
+    e = SqlStmt() { statements.add(e); }
```
dasch-1, line 4710: Can we use SqlStmtList() for this, which says that a SEMICOLON is optional after the last statement?

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ParserTest.java
```diff
@@ -4232,6 +4232,24 @@ private void checkExNotNull(SqlParserUtil.StringAndPos sap, Throwable thrown) {
     sql(sql).ok(expected);
   }
 
+  @Test public void testRenameProcedureTo() {
```
dasch-1, line 4235: Are these tests meant to be part of this PR?

michaelvdow: I think they got rearranged when I merged master, I can move them back down so it doesn't say it's added

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: It looks like it would parse correctly but it returns a different type of SqlNodeList instead of SqlStatementList. I think if we wanted to update that function it would require adding an import in every config file for SqlStatementList then updating the tests that use this function?

PR #257, opened 2020-08-03
[CALCITE] Handle ambiguous column resolution for unknown tables.

dstekol:
Modifies column resolution for unknown columns, so that unknown columns are not attributed to arbitrary unknown tables.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
@@ -216,8 +216,9 @@ public abstract class ListScope extends DelegatingScope {
         found++;
         if (child.namespace.getType() instanceof UnknownRecordType) {
           unknownFound++;
-          // If table is unknown, only set {@code type} to ANY if it has not already been set.
-          // This ensures that the column will be resolved to a known (non-ANY) type if possible.
+          // If table is unknown, only set {@code type} to ANY if it has not
+          // already been set. This ensures that the column will be resolved to
+          // a known (non-ANY) type if possible.
           if (type == null) {
             type = field.getType();
           }
@@ -232,11 +233,14 @@ public abstract class ListScope extends DelegatingScope {
     case 1:
       return type;
     default:
-      // Only return type if unknown tables are allowed, and at most one of the matched tables
-      // is a known table (if more than one known table is matched, then the column would be
-      // ambiguous even with full schema information). If more than one unknown table is matched,
-      // it doesn't matter which one the column actually comes from, since the column type is ANY.
-      if (validator.config().allowUnknownTables() && found - unknownFound <= 1) {
+      // Only return type if unknown tables are allowed, and at most one of the
+      // matched tables is a known table (if more than one known table is
+      // matched, then the column would be ambiguous even with full schema
+      // information). If more than one unknown table is matched, it doesn't
+      // matter which one the column actually comes from, since the column type
+      // is ANY.
+      if (validator.config().allowUnknownTables()
+          && found - unknownFound <= 1) {
         return type;
       }
       throw validator.newValidationError(ctx,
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
```diff
@@ -212,7 +214,16 @@ public RelDataType resolveColumn(String columnName, SqlNode ctx) {
           nameMatcher.field(childRowType, columnName);
       if (field != null) {
         found++;
-        type = field.getType();
+        if (child.namespace.getType() instanceof UnknownRecordType) {
+          unknownFound++;
+          // If table is unknown, only set {@code type} to ANY if it has not already been set.
```
dasch-1, line 219: nit: Line wrap

core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
```diff
@@ -212,7 +214,16 @@ public RelDataType resolveColumn(String columnName, SqlNode ctx) {
...
+          // This ensures that the column will be resolved to a known (non-ANY) type if possible.
```
dasch-1, line 220: nit: Line wrap

core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
```diff
@@ -221,9 +232,15 @@ public RelDataType resolveColumn(String columnName, SqlNode ctx) {
     case 1:
       return type;
     default:
+      // Only return type if unknown tables are allowed, and at most one of the matched tables
```
dasch-1, line 235: nit: Line wrap at 80 characters for this comment line and the rest in this block.

core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
```diff
@@ -221,9 +232,15 @@ public RelDataType resolveColumn(String columnName, SqlNode ctx) {
...
+      // is a known table (if more than one known table is matched, then the column would be
+      // ambiguous even with full schema information). If more than one unknown table is matched,
+      // it doesn't matter which one the column actually comes from, since the column type is ANY.
+      if (validator.config().allowUnknownTables() && found - unknownFound <= 1) {
```
dasch-1, line 239: nit: Line wrap

dasch-1, line 239: Wouldn't this match if, say, found == 2 and unknownFound == 2? Do we want that to match?

dstekol: I think we would - that would mean that both of the matched tables are unknown, which is fine. The only case we want to throw an error is when more than one known table is matched.

core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
dstekol: Oops, accidentally set wrapping to 100.

core/src/main/java/org/apache/calcite/sql/validate/ListScope.java
dasch-1: SGTM

core/src/test/java/org/apache/calcite/test/SqlValidatorBestEffortTest.java
```diff
@@ -143,4 +143,51 @@
...
+        + " ON `EMP`.`EMPNO` = `FOO`.`X`";
+    sql(sql)
+        .withValidatorIdentifierExpansion(true)
+        .rewritesTo(expected);
+  }
+
+  @Test public void testColumnFromKnownTableExpanded() {
+    String sql = "SELECT empno "
+        + "FROM foo "
+        + "INNER JOIN emp on emp.empno = foo.x";
```
dasch-1, line 165: Out of curiosity, where are "emp" and "dept" defined as known tables?

dstekol: [MockCatalogReaderSimple](https://github.com/dstekol/calcite/blob/paren_from_clause/core/src/test/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java)
Also linked in design doc in test section.

dasch-1: Great thanks

PR #258, opened 2020-08-03
[CALCITE] Added support for ALLOCATE cursor call in CREATE PROCEDURE

michaelvdow:
CursorStmt() method is setup to accept other cursor statements that will be added in later PRs.


core/src/main/java/org/apache/calcite/sql/SqlAllocateCursor.java
```diff
@@ -0,0 +1,63 @@
...
+      new SqlSpecialOperator("ALLOCATE CURSOR", SqlKind.ALLOCATE_CURSOR);
+
+  public final SqlIdentifier cursorName;
+  public final SqlIdentifier procedureName;
+
+  /**
+   * Creates a {@code SqlAllocateCursor}.
+   *
+   * @param pos  Parser position, must not be null
+   * @param cursorName  Name of the cursor previously opened
```
dasch-1, line 39: nit: Add "must not be null to comment"?

core/src/main/java/org/apache/calcite/sql/SqlAllocateCursor.java
```diff
@@ -0,0 +1,63 @@
...
+   * @param procedureName  Name of the procedure called
```
dasch-1, line 40: nit: Add "must not be null to comment"?

core/src/main/java/org/apache/calcite/sql/SqlAllocateCursor.java
michaelvdow: Sure I'll update these two comments

core/src/main/java/org/apache/calcite/sql/SqlAllocateCursor.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(cursorName, procedureName);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("ALLOCATE");
+    cursorName.unparse(writer, 0, 0);
+    writer.keyword("CURSOR FOR PROCEDURE");
+    procedureName.unparse(writer, 0, 0);
```
dstekol, line 61: Is there a reason you are passing in zeros instead of the given leftPrec and rightPrec parameters. Honestly not too familiar with this so just curious.

michaelvdow: That's how I've usually seen the unparse method used in SqlCall's although I'm not entirely sure why or what the leftPrec/rightPrec variables are used for.

dstekol: Fair enough. May be worth having someone look into the whole leftPrec/rightPrec thing in general at some point.

PR #261, opened 2020-08-03
[CALCITE] Add support for CLOSE cursor call in CREATE PROCEDURE statement 

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlCloseCursor.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code SqlCloseCursor} call.
+ */
+public class SqlCloseCursor extends SqlCall {
+  private static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("CLOSE", SqlKind.CLOSE);
```
dasch-1, line 30: Should this be something like SqlKind.CLOSE_CURSOR (same for the text string)?

core/src/main/java/org/apache/calcite/sql/SqlCloseCursor.java
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param pos SQL parser position
+   * @param cursorName Name of the cursor to close, must not be null
+   */
+  public SqlCloseCursor(SqlParserPos pos, SqlIdentifier cursorName) {
+    super(pos);
+    this.cursorName = Objects.requireNonNull(cursorName);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
```
dasch-1, line 45: nit: Line wrap.

core/src/main/java/org/apache/calcite/sql/SqlCloseCursor.java
michaelvdow: Yeah I think that would be clearer, will update

core/src/main/java/org/apache/calcite/sql/SqlCloseCursor.java
michaelvdow: In my editor it looks like this line is just at 80 chars.

dasch-1: You're right - looks good.

PR #262, opened 2020-08-03
[CALCITE] Add support for DEALLOCATE PREPARE call in CREATE PROCEDURE statement

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlDeallocatePrepare.java
```diff
@@ -0,0 +1,57 @@
...
+  private static final SqlSpecialOperator OPERATOR =
+      new SqlSpecialOperator("DEALLOCATE PREPARE", SqlKind.DEALLOCATE_PREPARE);
+
+  public final SqlIdentifier statementName;
+
+  /**
+   * Creates a {@code SqlDeallocatePrepare}.
+   *
+   * @param pos  Parser position, must not be null
+   * @param statementName  Name of the statement to deallocate, must not be null
```
dasch-1, line 38: nit: Line wrap

parsing/dialects/intermediate/dialects/dialect1/config.fmpp
```diff
@@ -77,6 +77,7 @@ data: {
       "org.apache.calcite.sql.SqlCreateTableDialect1",
       "org.apache.calcite.sql.SqlDateTimeAtLocal",
       "org.apache.calcite.sql.SqlDateTimeAtTimeZone",
+      "org.apache.calcite.sql.SqlDeallocatePrepare",
```
dasch-1, line 80: nit: Put in alphabetical order (e.g. after SqlDdlNodes). Also, can you remove the duplicate SqlDdlNodes while you're here?

michaelvdow: Yeah I'll update that

core/src/main/java/org/apache/calcite/sql/SqlDeallocatePrepare.java
michaelvdow: It looks like this line is also just at 80 chars. Is it supposed to be < 80 or is <=80 fine?

dasch-1: You're right - again! I just realized the helper utility I'm using to determine line lengths in these PRs had an off-by-one error. Should be fixed now.

PR #266, opened 2020-08-04
[CALCITE] Add support for EXECUTE statement in CREATE PROCEDURE statement

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlExecuteStatement.java
```diff
@@ -0,0 +1,67 @@
...
+  public final SqlNodeList parameters;
+
+  /**
+   * Creates a {@code SqlExecuteStatement}.
+   *
+   * @param pos  Parser position, must not be null
+   * @param statementName  Name of prepared statement to execute, must not be
+   *                       null
+   * @param parameters  List of parameters after USING keyword, must not be
+   *                       null
```
dasch-1, line 42: nit: Indentation not aligned with line above.

PR #267, opened 2020-08-04
[CALCITE] Fixed EXCEPT unparsing bug

AndrewPochapsky:
- Removed overriding of SqlSelect in BigQuery
- Override SelectExpression() in BigQuery

### DIFF
--- a/parsing/dialects/intermediate/dialects/bigquery/parserImpls.ftl
+++ b/parsing/dialects/intermediate/dialects/bigquery/parserImpls.ftl
@@ -15,94 +15,37 @@
 // limitations under the License.
 -->
 
-SqlExcept ExceptExpression(List<SqlNode> selectList):
+SqlExcept ExceptExpression():
 {
     final Pair<SqlNodeList, SqlNodeList> nameAndTypePair;
     final SqlNodeList exceptList;
 }
 {
-    <EXCEPT> nameAndTypePair = ParenthesizedCompoundIdentifierList()
+    <STAR> <EXCEPT> nameAndTypePair = ParenthesizedCompoundIdentifierList()
     {
         exceptList = nameAndTypePair.getKey();
-        if (selectList.size() > 0) {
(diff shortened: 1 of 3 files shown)


parsing/dialects/intermediate/dialects/bigquery/parserImpls.ftl
```diff
@@ -15,94 +15,37 @@
...
-    final SqlNodeList windowDecls;
-    final List<SqlNode> hints = new ArrayList<SqlNode>();
-    final Span s;
+    SqlNode e;
 }
 {
-    <SELECT>
-    {
-        s = span();
+    LOOKAHEAD(2)
```
dasch-1, line 39: @michaelvdow Isn't 2 the default LOOKAHEAD value?

AndrewPochapsky: I think it used to (still might be) but the code didn't work without this.

michaelvdow: Yeah, it looks like it's still set as 2 in the [build script](https://github.com/googleinterns/calcite/blob/master/parsing/dialects/intermediate/dialects/dialect1/build.gradle.kts#L62)

AndrewPochapsky: Looks like it is actually set to 1 for bigquery: https://github.com/googleinterns/calcite/blob/master/parsing/dialects/intermediate/dialects/bigquery/build.gradle.kts#L62

dasch-1: Aha, good catch. Forgot that it's set to 1 for all others except Dialect1.

michaelvdow: Ah okay, I thought this was for Dialect1 but that makes sense

PR #274, opened 2020-08-05
[CALCITE] Add support for SELECT AND CONSUME statement in BEGIN...END clause

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
```diff
@@ -0,0 +1,79 @@
...
+  public final SqlIdentifier fromTable;
+
+  /**
+   * Creates a {@code SqlSelectAndConsume}.
+   *
+   * @param pos  Parser position, must not be null
+   * @param set  True if the SET keyword was specified
+   * @param selectList  An asterisk or a list of SQL expressions, must not be
+   *                    null
+   * @param parameters A list of parameter names
```
dasch-1, line 45: nit: The above parameters have two spaces between the name and the description. Whereas this parameter and the next one just has a single space.

core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(selectList, parameters, fromTable);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("SELECT");
+    if (set) {
+      writer.keyword("SET");
+    }
+    writer.keyword("AND CONSUME TOP 1");
```
dasch-1, line 71: nit: Add a comment clarifying why this is hardcoded as "TOP 1".

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5049,3 +5051,48 @@ SqlCloseCursor SqlCloseCursor() :
...
+    boolean set = false;
+    final List<SqlNode> selectList;
+    final SqlNodeList parameters = new SqlNodeList(getPos());
+    final SqlIdentifier fromTable;
+    final int topNum;
+    final Span s = Span.of();
+    SqlNode e;
+}
+{
+    <SELECT>
```
dasch-1, line 5066: The docs state that <SEL> is also valid here.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5049,3 +5051,48 @@ SqlCloseCursor SqlCloseCursor() :
...
+        <COMMA>
+        (
+            e = SimpleIdentifier()
+        |
+            e = SqlHostVariable()
+        )
+        { parameters.add(e); }
+    )*
+    <FROM>
+    fromTable = SimpleIdentifier()
```
dasch-1, line 5092: Can this be a CompoundIdentifier?

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: Yeah you're right, I was looking at a newer version of the docs and I think it has a typo? It says "SET" is allowed instead of "SEL", but I think SEL would make more sense here

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: The docs don't explicitly say it but I think it can be so I'll update it.

core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
michaelvdow: Updated

core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
michaelvdow: Updated

core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
dstekol: Since this explicitly states that an error should be thrown if the query contains anything other than TOP 1, it might be good to also add a failing test case asserting this.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5063,3 +5065,46 @@ SqlCloseCursor SqlCloseCursor() :
     <CLOSE> cursorName = SimpleIdentifier()
     { return new SqlCloseCursor(s.end(this), cursorName); }
 }
+
+SqlSelectAndConsume SqlSelectAndConsume() :
```
dstekol, line 5069: Would it ever be valid to call this outside of CursorStmt()? If not, I think it would make sense to add a top-level comment to this function clarifying in which contexts this type of syntax is valid.

core/src/main/java/org/apache/calcite/sql/SqlSelectAndConsume.java
michaelvdow: Yeah that'd be a good test case, I'll add it in

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
michaelvdow: I was looking at the docs again and there is a form of SELECT AND CONSUME that can be used standalone but it doesn't allow the INTO clause. I could add a comment here clarifying this SELECT format is only used inside stored procedures

dasch-1: SGTM.

PR #275, opened 2020-08-05
[CALCITE] Moved Separators, Operators, and Identifiers from Parser.jj into plaintext files

AndrewPochapsky:
Builds on top of https://github.com/googleinterns/calcite/pull/268
- Updated parserTest/ directory to also test identifiers.txt
- Refactored how DialectGenerate unparses token assignment statements
   - Updated existing tests to work with this change
- Set up directory structure in parsing/ such that the change in https://github.com/googleinterns/calcite/pull/252 only applies to Dialect1
- Moved the operator and separator tokens in Parser.jj into parsing/keywords.txt

With https://github.com/googleinterns/calcite/pull/276 the operators.txt and separators.txt files are now unnecessary as no tests now check the full error message.


buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -214,6 +228,27 @@ private void traverse(
...
+  /**
+   * Adds all of the elements in {@code otherMap} to {@code mainMap}. If a key
+   * is already present in {@code mainMap} it is removed before being updated.
+   * This is done to ensure that the {@code filePath} gets updated if a given keyword
+   * has been overridden.
+   *
+   * @param mainMap The map that is getting entries added to it
+   * @param otherMap The map whose entries are being added from
+   */
+  private void addOrReplaceEntries(Map<Keyword, String> mainMap,
```
dasch-1, line 240: Can this function be marked as "private static"?

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -214,6 +228,27 @@ private void traverse(
...
+      Map<Keyword, String> otherMap) {
```
dasch-1, line 241: nit: Let's mark otherMap as final.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
dasch-1, line 240: Can we just use mainMap.putAll(otherMap) for this?

AndrewPochapsky: That was the original implementation but since filePath isn't used to compute hashCode or equals() only the value of the keyword would get replaced leaving the filePath to be not updated.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
```diff
@@ -214,6 +228,27 @@ private void traverse(
...
+   * @param mainMap The map that is getting entries added to it
+   * @param otherMap The map whose entries are being added from
+   */
+  private static void addOrReplaceEntries(Map<Keyword, String> mainMap,
+      final Map<Keyword, String> otherMap) {
+    for (Map.Entry<Keyword, String> entry : otherMap.entrySet()) {
+      Keyword key = entry.getKey();
+      String value = entry.getValue();
+      if (mainMap.containsKey(key)) {
+        mainMap.remove(key);
```
dstekol, line 246: Is this removal needed? My understanding is that put() overwrites any current entries with the given key.

AndrewPochapsky: Unfortunately it is required. This is because put() doesn't override the key but only the value. In this case we also want to override the key since the keyword's filePath is being updated.

dstekol: Ok, SGTM

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/DialectTraverser.java
dasch-1: Thanks for pointing this out. This might point to a larger issue in our usage of Keywords as keys.

Since Keyword doesn't implement hashCode() or equals() using all of its fields, it may not be a good contender as a key in a Set/Map. Keys in these data structures should match the intuition that if key1.equals(key2) is true, then the keys are identical.

AndrewPochapsky: I created a possible alternative here: https://github.com/AndrewPochapsky/calcite/pull/1. I changed Map<Keyword, String> to Map<String, Keyword> with a key of keyword name. Let me know what you think :)

dasch-1: Cool thanks for coding that up. LGTM.

AndrewPochapsky: Great, so I should merge it into this branch?

dasch-1: Yes, please do. And then we can review the version with those changes in this PR.

buildSrc/subprojects/parser/src/main/java/org/apache/calcite/buildtools/parser/Keyword.java
```diff
@@ -24,34 +24,50 @@
...
    */
-  public Keyword(String keyword, String filePath) {
+  public Keyword(String keyword, String value, String filePath) {
     this.keyword = Objects.requireNonNull(keyword.toUpperCase());
+    this.value = Objects.requireNonNull(value);
     this.filePath = filePath;
   }
 
   @Override public int hashCode() {
-    // The filePath should not be considered when calculating hashCode.
```
dstekol, line 45: Out of interest, what is the benefit of overriding the default hashCode function?

AndrewPochapsky: I believe that the benefit is that Keyword objects will now be more efficiently used within a hashing datastructure (such as HashMap).

dstekol: More efficiently than the default hashcode?

dstekol: Oh, nvm, I think I understand. You are overriding the equals method as well, so I guess the default hashcode wouldn't work for this.

PR #279, opened 2020-08-05
[CALCITE] Added support for UPDATE using cursor in BEGIN...END clause

michaelvdow:
- Added a test case for UPDATE and EXECUTE macro to make sure they are not being parsed by CursorStmt() if they appear in a CREATE PROCEDURE statement


core/src/main/java/org/apache/calcite/sql/SqlUpdateUsingCursor.java
```diff
@@ -0,0 +1,76 @@
...
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code SqlUpdateCursor} call.
```
dasch-1, line 26: nit: `{@code SqlUpdateUsingCursor}`

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -4640,6 +4640,10 @@ SqlNode CreateProcedureStmt() :
     (
         e = ConditionalStmt()
     |
+        // LOOKAHEAD ensure statements such as UPDATE table and EXECUTE macro
```
dasch-1, line 4643: nit: /ensure/ensures/

PR #281, opened 2020-08-05
[CALCITE] Disable star expansion for unknown tables.

dstekol:
Ensures that the select lists of "SELECT * " queries are not expanded when selecting from unknown tables.


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4310,6 +4310,16 @@ protected RelDataType validateSelectList(
     return typeFactory.createStructType(fieldList);
   }
 
+  private boolean containsDynamicStar(
+      List<Map.Entry<String, RelDataType>> fieldList) {
+    for (Map.Entry<String, RelDataType> item : fieldList) {
+      if (item.getValue().getSqlTypeName().getName().equals("DYNAMIC_STAR")) {
```
dasch-1, line 4316: Can we just compare getSqlTypeName() to the enum value of DYNAMIC_STAR rather than a string comparison?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4310,6 +4310,16 @@ protected RelDataType validateSelectList(
     return typeFactory.createStructType(fieldList);
   }
 
+  private boolean containsDynamicStar(
+      List<Map.Entry<String, RelDataType>> fieldList) {
```
dasch-1, line 4314: nit: Let's mark the fieldList as final.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4310,6 +4310,16 @@ protected RelDataType validateSelectList(
     return typeFactory.createStructType(fieldList);
   }
 
+  private boolean containsDynamicStar(
```
dasch-1, line 4313: nit: Can this class be "private static"?

PR #284, opened 2020-08-06
[CALCITE] Added support for OPEN cursor statement in BEGIN...END clause

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlOpenCursor.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Creates a {@code SqlOpenCursor}.
+   *
+   * @param pos  Parser position, must not be null
+   * @param cursorName  Name of the cursor to open, must not be null
+   * @param parameters  List of variables used as input to SQL statement, must
+   *                    not be null
+   */
+  public SqlOpenCursor(SqlParserPos pos, SqlIdentifier cursorName,
```
ArthurInTheShell, line 43: nit: add final keywords to all parameters?

michaelvdow: I'm not sure if I've seen final used in constructors except in a couple classes in Calcite. If others think this could be useful too it can be added in going forward since this would affect other classes being created as well

ArthurInTheShell: SGTM

PR #286, opened 2020-08-06
[CALCITE] Parsing INTERVALs that aren’t specified in strings

AndrewPochapsky:
Overrode IntervalLiteral() in Dialect1 and added the option for the literal to be an IntLiteral() in addition to the existing <QUOTED_STRING>


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5167,3 +5167,32 @@ SqlDeclareCursor SqlDeclareCursor() :
             statementName, preparedStatementName, prepareFrom);
     }
 }
+
+SqlLiteral IntervalLiteral() :
+{
+    final String p;
```
ArthurInTheShell, line 5173: nit: A more descriptive variable name might be useful here.

PR #287, opened 2020-08-06
[CALCITE] Iteration Statement: LOOP Statement and REPEAT Statement

ArthurInTheShell:
Add support for loop statement and repeat statement (Scripting) to calcite.


core/src/main/java/org/apache/calcite/sql/SqlIterationStmt.java
```diff
@@ -58,7 +58,7 @@ protected SqlIterationStmt(final SqlParserPos pos, final SqlNode condition,
       throw SqlUtil.newContextException(beginLabel.getParserPosition(),
           RESOURCE.beginEndLabelMismatch());
     }
-    this.condition = Objects.requireNonNull(condition);
+    this.condition = condition;
```
dasch-1, line 61: nit: Update the comment to match that this can be null

core/src/main/java/org/apache/calcite/sql/SqlLoopStmt.java
```diff
@@ -0,0 +1,57 @@
...
+    super(pos, /*condition = */null, statements, beginLabel, endLabel);
+  }
+
+  @Override public SqlOperator getOperator() {
+    return OPERATOR;
+  }
+
+  @Override public void unparse(final SqlWriter writer, final int leftPrec,
+      final int rightPrec) {
+    if (label != null) {
```
dasch-1, line 46: Should we move the label printing into the SqlIterationStmt class' unparse() method since it's not LOOP specific?

core/src/main/java/org/apache/calcite/sql/SqlIterationStmt.java
ArthurInTheShell: Good catch, fixed in https://github.com/googleinterns/calcite/pull/287/commits/5ec39d0522cb6f9f398b68a2482a2f18ed2ad4fc

core/src/main/java/org/apache/calcite/sql/SqlLoopStmt.java
ArthurInTheShell: I believe it would be good to abstract this behavior out into SqlIterationStmt class. But since the label printing is on **both** begin and end of the unparse method, I did not came up with one straight forward answer by utilizing the unparse method of the super class. However, I believe setting up two methods "preunparse" and "postunparse" in the super class can help making the order correct.

ArthurInTheShell: Changes made in https://github.com/googleinterns/calcite/pull/287/commits/1a806f61d827153ee8f4ad99cac5fe8424ef0834, please let me know what you think about it.

core/src/main/java/org/apache/calcite/sql/SqlIterationStmt.java
```diff
@@ -40,7 +40,8 @@
    * Creates a {@code SqlIterationStmt}.
    *
    * @param pos         Parser position, must not be null.
-   * @param condition   Conditional expression, must not be null.
+   * @param condition   Conditional expression, can be null for loop statement
```
dasch-1, line 43: nit: This base class should have no information/context about its subclasses. Suggest simplifying this comment to just "Conditional expression."

ArthurInTheShell: Make sense. Updated in https://github.com/googleinterns/calcite/pull/287/commits/ee4958b16cdd8fb04f9d6c925b4ccf26767b6cc6

PR #292, opened 2020-08-07
[CALCITE] Added support for DECLARE HANDLER statements in BEGIN...END clause

michaelvdow:
Syntax:
`DECLARE ((CONTINUE | EXIT) HANDLER | <condition_name> CONDITION) [FOR (<sql_state> (, <sql_state>)* [<handler_statement>] | <condition_type> (, <condition_type>)* <handler_statement>)]`

`<sql_state>` = `SQLSTATE [VALUE] <code>`

`<condition_type>` = `(SQLEXCEPTION | SQLWARNING | NOT FOUND | <condition_name>)`


core/src/main/java/org/apache/calcite/sql/SqlState.java
```diff
@@ -0,0 +1,40 @@
...
+   * @param value A SQLSTATE value, must not be null
+   * @param pos Parser position, must not be null
+   */
+  public SqlState(String value, SqlParserPos pos) {
+    super(value, pos);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("SQLSTATE");
+    writer.literal("'" + names.get(0) + "'");
```
dasch-1, line 38: Can we use getSimple() instead of names.get(0)?

core/src/main/java/org/apache/calcite/sql/SqlDeclareHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(conditionName, parameters,
+        handlerStatement);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("DECLARE");
+    switch (handlerType) {
+    case CONTINUE:
```
lisayin, line 70: nit: indent case statements

dasch-1: Drive-by comment: the style used in this code base is to not indent the case beyond the switch.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5304,3 +5310,85 @@ SqlWhileStmt WhileStmt() :
...
+                RESOURCE.sqlStateCharLength(value));
+        }
+    }
+    { return new SqlState(value, s.end(this)); }
+}
+
+SqlIdentifier DeclareHandlerCondition() :
+{
+    final Span s = Span.of();
+    DeclareHandlerConditionType conditionType;
```
dstekol, line 5377: nit: can this be declared as final?

PR #295, opened 2020-08-07
[CALCITE] Iteration Statement: FOR Statement

ArthurInTheShell:
Added support for For Statement with tests.


core/src/main/java/org/apache/calcite/sql/SqlForStmt.java
```diff
@@ -0,0 +1,83 @@
...
+  public final SqlNode cursorSpecification;
+
+  /**
+   * Creates a {@code SqlForStmt}.
+   * @param pos         Parser position, must not be null.
+   * @param statements  List of statements to iterate, must not be null.
+   * @param beginLabel  Optional begin label, must match end label if not null.
+   * @param endLabel    Optional end label, must match begin label if not null.
+   * @param forLoopVariable
+   *                    The name of the loop.
```
dasch-1, line 43: nit: Should this be "loop variable"?

dasch-1: Also, should we say that it "must not be null"?

core/src/main/java/org/apache/calcite/sql/SqlForStmt.java
```diff
@@ -0,0 +1,83 @@
...
+   * @param cursorName  The name of the cursor.
+   * @param cursorSpecification
+   *                    A single select statement used as the cursor.
```
dasch-1, line 46: nit: "must not be null"?

core/src/main/java/org/apache/calcite/sql/SqlForStmt.java
ArthurInTheShell: Updated in https://github.com/googleinterns/calcite/pull/295/commits/0729b1498ee2b87bc6f1697d05000da7b7fb4228

core/src/main/java/org/apache/calcite/sql/SqlForStmt.java
ArthurInTheShell: Added in https://github.com/googleinterns/calcite/pull/295/commits/0729b1498ee2b87bc6f1697d05000da7b7fb4228

PR #296, opened 2020-08-07
[CALCITE] Add support for SELECT INTO statement in BEGIN...END clause

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlSelectInto.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override public List<SqlNode> getOperandList() {
+    return ImmutableNullableList.of(selectList, parameters, fromClause,
+        whereClause);
+  }
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    writer.keyword("SELECT");
+    if (selectKeyword != null) {
```
dasch-1, line 71: nit: Why not use the "UNSPECIFIED" pattern used elsewhere instead of checking for null?

michaelvdow: I was using an existing enum class used for SELECT that didn't contain UNSPECIFIED, but I could update it to use that pattern here.

core/src/main/java/org/apache/calcite/sql/SqlSelectInto.java
```diff
@@ -0,0 +1,92 @@
...
+   *                    null
+   * @param parameters A list of parameter names, must not be null
+   * @param fromClause FROM clause
+   * @param whereClause WHERE clause
+   */
+  public SqlSelectInto(SqlParserPos pos, SqlSelectKeyword selectKeyword,
+      SqlNodeList selectList, SqlNodeList parameters, SqlNode fromClause,
+      SqlNode whereClause) {
+    super(pos);
+    this.selectKeyword = selectKeyword;
```
dasch-1, line 53: nit: Can we add Objects.requireNonNull() here now? And update the Javadoc to reflect?

PR #299, opened 2020-08-10
[CALCITE] FIRST_VALUE/LAST_VALUE with IGNORE/RESPECT NULLS

AndrewPochapsky:
- Added AST node SqlFirstLastValue
   - This doesn't have a field for the window because that is unparsed in a somewhat special way
- Added parsing logic
- Added tests


core/src/main/java/org/apache/calcite/sql/SqlFirstLastValue.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+    if (first) {
+      writer.keyword("FIRST_VALUE");
+    } else {
+      writer.keyword("LAST_VALUE");
+    }
+    SqlWriter.Frame frame = writer.startList("(", ")");
+    value.unparse(writer, leftPrec, rightPrec);
+    if (nullOption != null) {
```
dasch-1, line 69: nit: Why not use an UNSPECIFIED enum value here to be more explicit?

core/src/main/java/org/apache/calcite/sql/SqlFirstLastValue.java
```diff
@@ -0,0 +1,95 @@
...
+   * @param first  Whether or not this is the FIRST_VALUE or LAST_VALUE
+   * @param value  The value which is being used
+   * @param nullOption  If the nulls should be ignored or respected
+   */
+  public SqlFirstLastValue(SqlParserPos pos, boolean first, SqlNode value,
+      NullOption nullOption) {
+    super(pos);
+    this.first = first;
+    this.value = Objects.requireNonNull(value);
+    this.nullOption = nullOption;
```
dasch-1, line 50: Objects.requireNonNull()?

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -547,6 +547,9 @@
    */
   VALUES,
 
+  /** {@code FIRST_LAST_VALUE} expression. */
+  FIRST_LAST_VALUE,
```
ArthurInTheShell, line 551: I am not sure if it is necessary for this expression to have its own SqlKind, would it be in either SqlKind.OTHER or SqlKind.OTHER_FUNCTION? Could you provide more context on why FIRST_LAST_VALUE should have it's own SqlKind?

AndrewPochapsky: Having a specific SqlKind may be preferable by downstream consumers. What do you think @dasch-1 ?

dasch-1: Yes, having a distinct SqlKind is always preferred by downstream consumers.

PR #300, opened 2020-08-10
[CALCITE] Allow optional IS prefix to IN operator

dasch-1:
[CALCITE] Allow optional IS prefix to IN operator

### DIFF
--- a/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
+++ b/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
@@ -3173,13 +3173,14 @@ List<Object> Expression2(ExprContext exprContext) :
             (
                 // Special case for "IN", because RHS of "IN" is the only place
                 // that an expression-list is allowed ("exp IN (exp1, exp2)").
-                LOOKAHEAD(2) {
+                LOOKAHEAD(3) {
                     checkNonQueryExpression(exprContext);
                 }
                 (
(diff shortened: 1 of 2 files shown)


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3173,13 +3173,14 @@ List<Object> Expression2(ExprContext exprContext) :
...
+                LOOKAHEAD(3) {
                     checkNonQueryExpression(exprContext);
                 }
                 (
-                    ( <NOT> | <CARET> ) <IN> { op = SqlStdOperatorTable.NOT_IN; }
+                    [ <IS> ] ( <NOT> | <CARET> ) <IN>
+                    { op = SqlStdOperatorTable.NOT_IN; }
                 |
-                    <IN> { op = SqlStdOperatorTable.IN; }
+                    [ <IS> ] <IN> { op = SqlStdOperatorTable.IN; }
```
AndrewPochapsky, line 3183: Should we also add an optional <IS> to here? https://github.com/googleinterns/calcite/blob/d8b5d8102386f1ea37a543073b1a220fb8e904f2/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl#L3758

dasch-1: That block of code handles the parsing of the built-in POSITION function. AFAICT from the docs, that does not allow the IS prefix.

PR #301, opened 2020-08-10
[CALCITE] Add support for set statement in create procedure

ArthurInTheShell:


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5556,3 +5559,17 @@ SqlIterateStmt IterateStmt() :
...
+
+SqlSetStmt SetStmt() :
+{
+    final SqlIdentifier target;
+    final SqlNode source;
+    final Span s = Span.of();
+}
+{
+    <SET>
+    target = SimpleIdentifier()
```
AndrewPochapsky, line 5571: Can target be a CompoundIdentifier()?

ArthurInTheShell: It cannot be a CompoundIdentifier(), since it is supposed to be either a parameter or a local variable of the store procedure .

PR #304, opened 2020-08-11
[CALCITE] Added support for SIGNAL/RESIGNAL calls in CREATE PROCEDURE statements

michaelvdow:


core/src/main/java/org/apache/calcite/sql/SqlSignal.java
```diff
@@ -0,0 +1,81 @@
...
+   * @param pos Parser position, must not be null
+   * @param signalType Specifies if call is SIGNAL or RESIGNAL, must not be null
+   * @param conditionOrSqlState A condition name or SQLSTATE value
+   * @param setStmt The optional SET clause
+   */
+  public SqlSignal(SqlParserPos pos, SignalType signalType,
+      SqlNode conditionOrSqlState, SqlSetStmt setStmt) {
+    super(pos);
+    this.signalType = Objects.requireNonNull(signalType);
+    this.conditionOrSqlState = conditionOrSqlState;
```
dasch-1, line 48: nit: Use the UNSPECIFIED pattern and check that this is non null.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5514,6 +5514,43 @@ SqlNode DiagnosticStmt() :
...
+
+SqlNode SignalConditionOrSqlState() :
+{
+    final SqlNode e;
+}
+{
+    (
+        e = SqlState()
+    |
+        e = SimpleIdentifier()
```
dasch-1, line 5553: nit: Can these be ordered alphabetically?

michaelvdow: Yeah, I'll update those

core/src/main/java/org/apache/calcite/sql/SqlSignal.java
michaelvdow: Did you mean the signalType because the conditionOrSqlState is a SqlNode? There's already a non null check for signal type since those are the only two options

dasch-1: Ah, you're right. I meant the signalType, which does already have the requireNonNull check.

PR #307, opened 2020-08-12
[CALCITE] Redo of FIRST_VALUE/LAST_VALUE

AndrewPochapsky:
Redo of https://github.com/googleinterns/calcite/pull/299

- Created new AST SqlNullTreatment which holds preceding value and IGNORE/RESPECT NULLS
- For consistency, this AST is now also used for the existing SqlNullTreatmentOperator


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5551,6 +5553,58 @@ SqlGetDiagnosticsParam SqlGetDiagnosticsParam() :
...
+    final boolean first;
+    final SqlNode value;
+    final SqlNode over;
+    NullOption nullOption = NullOption.UNSPECIFIED;
+    final SqlCall firstLastCall;
+}
+{
+    (
+        <FIRST_VALUE> {
+            first = true;
```
ArthurInTheShell, line 5567: nit: Instead of using a boolean here, maybe we could define a final SqlAggFunc variable that hold either SqlStdOperatorTable.FIRST_VALUE or SqlStdOperatorTable.LAST_VALUE. So we can reduce the createCall() logic into one single line.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3979,6 +3979,8 @@ SqlNode BuiltinFunctionCall() :
         node = CaseN() { return node; }
     |
         node = RangeN() { return node; }
+    |
+        node = FirstLastValue() { return node; }
```
dasch-1, line 3983: Since we plan to parse FIRST_VALUE and LAST_VALUE using this function in Dialect1, let's remove "<FIRST_VALUE>" and "<LAST_VALUE>" from Dialect1's ReservedFunctionName() list.

core/src/main/java/org/apache/calcite/sql/SqlNullTreatment.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Parse tree for {@code SqlNullTreatment} statement.
+ */
+public class SqlNullTreatment extends SqlCall implements SqlExecutableStatement {
```
dasch-1, line 29: Why derive from a SqlCall here rather than a SqlNode?

PR #309, opened 2020-08-13
[CALCITE] FIRST LAST VALUE redone

AndrewPochapsky:
- Added new SqlNode to hold RESPECT/IGNORE NULLS
- Overrode unparse method of first last value function to unparse without comma
- Removed FIRST_VALUE and LAST_VALUE from dialect1's ReservedFunctionName()
- Removed old usage of RESPECT/IGNORE NULLS from Dialect1


parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5741,3 +5740,124 @@ SqlSelectInto SqlSelectInto() :
...
+    {
+        return SqlStdOperatorTable.OVER.createCall(getPos(),
+            firstLastCall, over);
+    }
+}
+
+/**
+ * Parses a reserved word which is used as the name of a function.
+ */
+SqlIdentifier ReservedFunctionName() :
```
ArthurInTheShell, line 5794: Is it safe to assume this new function is overriding a function in parsing/parserImpls.ftl? Maybe a piece of comment would help clarify.

AndrewPochapsky: You are correct, it is overriding. I'll add a comment.

core/src/main/java/org/apache/calcite/sql/SqlNullTreatmentModifier.java
```diff
@@ -0,0 +1,60 @@
...
+  public final SqlKind kind;
+
+  public SqlNullTreatmentModifier(SqlParserPos pos, SqlKind kind) {
+    super(pos);
+    this.kind = Objects.requireNonNull(kind);
+    Preconditions.checkArgument(kind == SqlKind.RESPECT_NULLS
+        || kind == SqlKind.IGNORE_NULLS);
+  }
+
+  @Override public SqlNode clone(SqlParserPos pos) {
```
ArthurInTheShell, line 40: nit: For method clone, we should either return null and add comment or return a new SqlNullTreatmentModifier as specified in SqlNode abstract class.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
dasch-1: A comment can't hurt. But I'm also not sure it's entirely necessary. We're already overloading lots of functions in Dialect1 that don't have comments (e.g. SqlSelect).

No strong preference from me on this.

AndrewPochapsky: I'll add a comment specifying that we removed FIRST_VALUE and LAST_VALUE, otherwise it may be difficult to tell what the difference is.

core/src/main/java/org/apache/calcite/sql/SqlNullTreatmentModifier.java
```diff
@@ -0,0 +1,60 @@
...
+      SqlValidator validator,
+      SqlValidatorScope scope) {
+  }
+
+  @Override public <R> R accept(SqlVisitor<R> visitor) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public boolean equalsDeep(SqlNode node, Litmus litmus) {
+    return false;
```
ArthurInTheShell, line 54: nit: A comment to explain if it is intentionally return false might be useful here.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
dasch-1: The issue with a comment like that is that it will immediately grow out of date (or be onerous to keep updated) as soon as we start adding/removing other keywords from here.

dasch-1: I think the comments/annotations added in https://github.com/googleinterns/calcite/pull/215 should be sufficient to help a developer understand what's being overloaded and what's not.

AndrewPochapsky: Ok, SGTM.

core/src/main/java/org/apache/calcite/sql/fun/SqlFirstLastValueAggFunction.java
```diff
@@ -78,4 +80,16 @@ public RelDataType getReturnType(RelDataTypeFactory typeFactory) {
...
     return true;
   }
+
+  @Override public void unparse(SqlWriter writer, SqlCall call, int leftPrec,
+      int rightPrec) {
+    writer.keyword(getName());
+    final SqlWriter.Frame frame =
+        writer.startList(SqlWriter.FrameTypeEnum.FUN_CALL, "(", ")");
+    call.operand(0).unparse(writer, leftPrec, rightPrec);
+    if (call.operandCount() == 2) {
```
dasch-1, line 90: nit: Suggest adding a comment describing what we're doing with this if-check since it may not be immediately obvious to a future reader.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5741,3 +5740,124 @@ SqlSelectInto SqlSelectInto() :
...
+    value = CompoundIdentifier() { args.add(value); }
+    [
+        (
+            <IGNORE> {
+                nullTreatmentModifier = new SqlNullTreatmentModifier(getPos(),
+                    SqlKind.IGNORE_NULLS);
+            }
+        |
+            <RESPECT> {
+                nullTreatmentModifier = new SqlNullTreatmentModifier(getPos(),
```
dasch-1, line 5773: nit: Since the call to the SqlNullTreatmentModifier constructor is common between these two cases, suggest doing that after the clause ends. Instead, we can just set the SqlKind value in the respective cases.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -5741,3 +5740,124 @@ SqlSelectInto SqlSelectInto() :
...
     }
 }
+
+SqlCall FirstLastValue() :
+{
+    final SqlNode value;
+    final SqlNode over;
+    final List<SqlNode> args = new ArrayList<SqlNode>();
+    SqlNullTreatmentModifier nullTreatmentModifier = null;
+    final SqlAggFunction function;
```
dasch-1, line 5750: Why use SqlAggFunction rather than SqlFirstLastValueAggFunction?

core/src/main/java/org/apache/calcite/sql/SqlNullTreatmentModifier.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.common.base.Preconditions;
+
+import java.util.Objects;
+
+/** Simple class to hold null treatment info. */
+public class SqlNullTreatmentModifier extends SqlNode {
+  public final SqlKind kind;
+
+  public SqlNullTreatmentModifier(SqlParserPos pos, SqlKind kind) {
```
dasch-1, line 33: nit: Missing javadoc specifying, for example, that "kind" should not be null.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
AndrewPochapsky: I used SqlAggFunction since that is the type which is returned by the FIRST_VALUE and LAST_VALUE calls: https://github.com/googleinterns/calcite/blob/67b7bb66aabf05a58171096356dc71476bf76381/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java#L977

dasch-1: SGTM

PR #310, opened 2020-08-14
[CALCITE] Fixed validation issue with caret negation operator

michaelvdow:


parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -28,4 +38,15 @@
         SqlTestFactory.INSTANCE
             .with("parserFactory", Dialect1ParserImpl.FACTORY));
   }
+
+  @Test public void testCaretNegation() throws SqlParseException {
```
dasch-1, line 42: We probably don't want to do all of this setup for each test case. I've moved some of this into a common location in: https://github.com/googleinterns/calcite/pull/311

Suggest waiting for that to be merged and then remove a lot of this setup logic.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -28,4 +38,15 @@
...
+    String sql = "select * from dept where ^deptno = 1";
+    SqlParser.Config config = configBuilder()
+        .setParserFactory(Dialect1ParserImpl.FACTORY)
+        .build();
+    SqlParser sqlParserReader = SqlParser.create(new StringReader(sql), config);
+    SqlNode node = sqlParserReader.parseQuery();
+    SqlValidator validator = tester.getValidator();
+    validator.validate(node);
```
dstekol, line 50: Would it be possible to call toSqlString() on the validated node similar to what is done [here](https://github.com/googleinterns/calcite/blob/aefe2b15dc4f187a740d2c4227d0f642320b6153/core/src/test/java/org/apache/calcite/sql/test/AbstractSqlTester.java#L505), and assert that it rewrites to what should be expected? Otherwise, the ^ could theoretically be dropped and we wouldn't know.

dasch-1: Completely agree. @dstekol are there benefits of using toSqlString() over rewritesTo(), as is done in the link below?

https://github.com/googleinterns/calcite/blob/aefe2b15dc4f187a740d2c4227d0f642320b6153/parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java#L42

dstekol: Michael mentioned in the design doc that ^ gets dropped by the toSql() call of the testing framework (since it usually gets used for annotating errors), so this is a workaround.

dstekol: I think that is also why this particular test sets everything up from scratch. We likely won't need to do this for anything else, but this is a special case (at least until someone figures out a way to indicate to the tester than ^ is being used as an operator).

michaelvdow: Yeah I initially set it up like that because the caret gets dropped, but I I think calling getTester().rewritesTo() would be a lot cleaner and that shouldn't remove the caret operator. If rewritesTo() is called on a sql() object the operator is removed.

michaelvdow: Yeah I initially set it up like that because the caret gets dropped, but I I think calling getTester().checkRewrite() which calls the toSqlString() method could be a lot cleaner and that shouldn't remove the caret operator. If rewritesTo() is called on a sql() object the operator is removed.

dasch-1: Ah I see. Let's at least comment that then.

michaelvdow: I updated the test case and left a comment. Thanks for pointing this out @dasch-1 and @dstekol.

PR #312, opened 2020-08-14
[CALCITE] Add validation support for host variables

dasch-1:


parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -41,4 +41,11 @@
         + "FROM `ABC` AS `ABC`";
     sql(sql).rewritesTo(expected);
   }
+
+  @Test public void testHostVariable() {
```
dstekol, line 45: Are there any restrictions we should be checking regarding where host variables can/cannot occur, what type they have, etc?

dasch-1: For now, I'd just like to be able to not alter host variables during validation rewriting. Actually extracting meaningful type information for them would require a larger scope effort (which may or may not be necessary given current needs).

dasch-1: The location where host variables can occur should already be enforced during parse time.

dstekol: Ok, SGTM.

PR #317, opened 2020-08-17
[CALCITE] Reference SqlLabeledBlock objects in SqlLeaveStmt through validation

michaelvdow:
- SqlScriptingNode was created to abstract the empty validate() method needed for scripting nodes
 - SqlLabeledBlock was created to abstract the label and statements variables from SqlBeginEndCall and SqlIterationStmt
 - BlockScope stores SqlLabeledBlock and resolve() method is setup to match the label names
 - LabeledBlockNamespace created since the Resolve object [requires](https://github.com/googleinterns/calcite/blob/67b7bb66aabf05a58171096356dc71476bf76381/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorScope.java#L329) a namespace
 - registerQuery() method in SqlValidatorImpl updated to register scopes and namespaces

Future PRs will build off of this to connect labels to ITERATE statements, as well as connecting condition declarations.


core/src/main/java/org/apache/calcite/sql/validate/LabeledBlockNamespace.java
```diff
@@ -0,0 +1,54 @@
...
+   * @param enclosingNode Enclosing node
+   */
+  LabeledBlockNamespace(SqlValidatorImpl validator,
+      SqlLabeledBlock labeledBlock,
+      SqlNode enclosingNode) {
+    super(validator, enclosingNode);
+    this.labeledBlock = Objects.requireNonNull(labeledBlock);
+  }
+
+  protected RelDataType validateImpl(RelDataType targetRowType) {
```
dasch-1, line 47: Missing `@Override` annotation

core/src/main/java/org/apache/calcite/sql/validate/LabeledBlockNamespace.java
```diff
@@ -0,0 +1,54 @@
...
+    return getValidator().getUnknownType();
+  }
+
+  public SqlNode getNode() {
```
dasch-1, line 51: Missing `@Override` annotation

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1167,6 +1172,10 @@ public SqlValidatorScope getJoinScope(SqlNode node) {
     return scopes.get(stripAs(node));
   }
 
+  public BlockScope getBlockScope(SqlLabeledBlock block) {
```
dasch-1, line 1175: Missing `@Override` annotation.

core/src/main/java/org/apache/calcite/sql/SqlLabeledBlock.java
```diff
@@ -0,0 +1,55 @@
...
+  public SqlLabeledBlock(SqlParserPos pos, SqlIdentifier label,
+      SqlStatementList statements) {
+    super(pos);
+    this.label = label;
+    this.statements = Objects.requireNonNull(statements);
+  }
+
+  @Override public void validate(final SqlValidator validator,
+      final SqlValidatorScope scope) {
+    SqlValidatorImpl validatorImpl = (SqlValidatorImpl) validator;
```
dasch-1, line 51: There's no need to cast to SqlValidatorImpl here since getBlockScope() is defined at the SqlValidator level too.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2880,11 +2889,74 @@ private void registerQuery(
...
+    case WHILE_STATEMENT:
+    case REPEAT_STATEMENT:
+    case LOOP_STATEMENT:
+      SqlLabeledBlock labeledBlock = (SqlLabeledBlock) node;
+      BlockScope blockScope = new BlockScope(parentScope, labeledBlock);
+      scopes.put(labeledBlock, blockScope);
+      LabeledBlockNamespace labeledBlockNs =
+          new LabeledBlockNamespace(this, labeledBlock, enclosingNode);
+      String blockAlias = deriveAlias(labeledBlock, nextGeneratedId++);
+      registerNamespace(blockScope, blockAlias, labeledBlockNs, false);
```
dasch-1, line 2910: nit: Use inline comment to indicate what parameter `false` represents.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2880,11 +2889,74 @@ private void registerQuery(
...
+      SqlIfStmt ifStmt = (SqlIfStmt) node;
+      registerStatementList(parentScope, usingScope, alias,
+          ifStmt.conditionalStmtListPairs, ifStmt);
+      registerStatementList(parentScope, usingScope, alias,
+          ifStmt.elseStmtList, ifStmt);
+      break;
+
+    case CASE_STATEMENT:
+      SqlCaseStmt caseStmt = (SqlCaseStmt) node;
+      registerStatementList(parentScope, usingScope, alias,
```
dasch-1, line 2925: It seems like IF and CASE share a lot of duplicate code. Suggest creating a base class (e.g. SqlConditional) that both derive from. And then implement all of the shared logic in one place.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2880,11 +2889,74 @@ private void registerQuery(
...
+    Set<SqlKind> supportedKinds = new HashSet<>(
+        Arrays.asList(SqlKind.BEGIN_END, SqlKind.WHILE_STATEMENT,
+            SqlKind.FOR_STATEMENT, SqlKind.REPEAT_STATEMENT,
+            SqlKind.LOOP_STATEMENT, SqlKind.DECLARE_CONDITION,
+            SqlKind.DECLARE_HANDLER, SqlKind.IF_STATEMENT,
+            SqlKind.CASE_STATEMENT, SqlKind.CONDITION_STATEMENT_LIST_PAIR));
+    for (SqlNode child : statements) {
+      if (supportedKinds.contains(child.getKind())) {
+        registerQuery(parentScope, usingScope, child,
+            enclosingNode, alias, false, false);
```
dasch-1, line 2955: nit: Use inline commenst to indicate what parameters `false, false` represent.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -101,4 +110,117 @@
         + "FROM `GHI` AS `GHI`";
     sql(sql).rewritesTo(expected);
   }
+
+  @Test public void testCreateProcedureBeginEndLabel() {
+    String sql = "create procedure foo()\n"
+        + "label1: begin\n"
+        + "leave label1;\n"
+        + "end";
+    SqlValidator validator = getTester().getValidator();
```
dasch-1, line 119: nit: Suggest putting lines 119 and 120 in a private helper method which just takes the `sql` as a parameter and returns a `SqlNode` (e.g. could also be named `parseAndValidate()`). Same for other tests too.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
michaelvdow: That's a good idea, I'll update it to move some logic out of IF and CASE.

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmt.java
```diff
@@ -0,0 +1,54 @@
...
+      SqlNodeList conditionalStmtListPairs, SqlNodeList elseStmtList) {
+    super(pos);
+    this.conditionalStmtListPairs =
+        Objects.requireNonNull(conditionalStmtListPairs);
+    this.elseStmtList = Objects.requireNonNull(elseStmtList);
+  }
+
+  @Override public void validate(SqlValidator validator,
+      SqlValidatorScope scope) {
+    validateStatementList(validator, scope, conditionalStmtListPairs);
```
ArthurInTheShell, line 51: Though it seems to me that we would need to validate conditionalStmtListPairs in a similar manner (iterate through and call validate on each of the element in the list) to validateStatementList, conditionalStmtListPairs is not a statement list. 
To prevent confusions, I would suggest we either add a comment here, or change the name(or javadoc) of validateStatementList to a more generalized one.

core/src/main/java/org/apache/calcite/sql/validate/BlockScope.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    super.resolve(names, nameMatcher, deep, resolved);
+  }
+
+  /**
+   * Searches this scope and parent scopes for a {@link SqlLabeledBlock} that
+   * matches the provided label.
+   *
+   * @param label The label of the block to find
+   * @return The labeled block
```
ArthurInTheShell, line 75:
```suggestion
   * @return The labeled block, return null if label is not found
```

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmt.java
michaelvdow: Sure I could change it to validateSqlNodeList() so it's more general

core/src/main/java/org/apache/calcite/sql/SqlConditionalStmt.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Creates a {@code SqlConditionalStmt}.
+   * @param pos                       Parser position, must not be null.
+   * @param conditionalStmtListPairs  List of conditional expression pairs
+   *                                  with StatementList, must not be null.
+   * @param elseStmtList              List of statements in the else clause,
+   *                                  must not be null.
+   */
+  public SqlConditionalStmt(SqlParserPos pos,
```
dasch-1, line 41: This constructor should be protected.

michaelvdow: Ah ok, I could make the SqlLabeledBlock constructor protected too

PR #318, opened 2020-08-17
[CALCITE] Extract schema for CREATE TABLE statements

AndrewPochapsky:
- Created new class ExplicitRowTypeTable that extends from Table
- Add validation logic to SqlCreateTable that creates an instance of ExplicitRowTypeTable and adds it to the schema
- Added tests

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.sql.validate;
 
 import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.jdbc.CalciteSchema;
 import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.linq4j.function.Function2;
 import org.apache.calcite.linq4j.function.Functions;
@@ -38,6 +39,7 @@ import org.apache.calcite.runtime.Feature;
 import org.apache.calcite.runtime.Resources;
 import org.apache.calcite.schema.ColumnStrategy;
 import org.apache.calcite.schema.Table;
+import org.apache.calcite.schema.impl.ExplicitRowTypeTable;
 import org.apache.calcite.schema.impl.ModifiableViewTable;
 import org.apache.calcite.sql.JoinConditionType;
 import org.apache.calcite.sql.JoinType;
@@ -45,9 +47,12 @@ import org.apache.calcite.sql.SqlAccessEnum;
 import org.apache.calcite.sql.SqlAccessType;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlBasicCall;
+import org.apache.calcite.sql.SqlBasicTypeNameSpec;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlColumnAttribute;
+import org.apache.calcite.sql.SqlColumnDeclaration;
+import org.apache.calcite.sql.SqlCreateTable;
 import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDelete;
 import org.apache.calcite.sql.SqlDynamicParam;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3219,6 +3227,34 @@ public void validateIntervalQualifier(SqlIntervalQualifier qualifier) {
     }
   }
 
+  @Override public void validateCreateTable(SqlCreateTable createTable) {
+    Preconditions.checkArgument(createTable != null);
+    // Type doesn't matter for CREATE TABLE statements.
```
ArthurInTheShell, line 3232: nit: doesn't matter/ is not applicable

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3219,6 +3227,34 @@ public void validateIntervalQualifier(SqlIntervalQualifier qualifier) {
...
+    nodeToTypeMap.put(createTable, unknownType);
+    CalciteSchema schema = catalogReader.getRootSchema();
+    RelDataTypeFactory.Builder builder = typeFactory.builder();
+    if (createTable.columnList == null) {
```
dasch-1, line 3236: Is it legal for the column list be null (not just empty, but null)? According to the Dialect1 docs, it doesn't look legal and I'm not sure of a dialect in which it would be.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -101,4 +101,32 @@
         + "FROM `GHI` AS `GHI`";
     sql(sql).rewritesTo(expected);
   }
+
+  @Test public void testCreateTableNoColumns() {
```
dasch-1, line 105: Not sure of omitting the columns list altogether is legal. Instead, should we just do a test case with a table with no columns?

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -101,4 +101,32 @@
...
+    String ddl = "create table foo";
+    String query = "select * from foo";
+    sql(ddl).ok();
+    sql(query).type("RecordType() NOT NULL");
+  }
+
+  @Test public void testCreateTableWithColumns() {
```
dasch-1, line 112: Suggest breaking all of these into their own individual test cases. Even though that results in duplicating the same setup code, it simplifies the testing logic and makes it easier for a future developer to reason about what caused a test to break.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I'm also unsure, I went off of the code of SqlCreateTable where it states that the column list "may be null", WDYT? https://github.com/googleinterns/calcite/blob/67b7bb66aabf05a58171096356dc71476bf76381/core/src/main/java/org/apache/calcite/sql/SqlCreateTable.java#L88. It also defaults to null if there is no column list here: https://github.com/googleinterns/calcite/blob/67b7bb66aabf05a58171096356dc71476bf76381/parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl#L936

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
AndrewPochapsky: Sure, SGTM.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -109,24 +109,39 @@
...
+    String ddl = "create table foo(x int, y varchar)";
+    String query = "insert into foo values (1, 'str')";
+    sql(ddl).ok();
+    sql(query).type("RecordType(INTEGER NOT NULL X, VARCHAR NOT NULL Y)"
         + " NOT NULL");
+  }
 
-    String query4 = "insert into foo values (1, 'str', 1)";
-    sql(query4).fails("end index \\(3\\) must not be greater than size \\(2\\)");
+  @Test public void testCreateTableSelectNonExistingColumnFails() {
```
dasch-1, line 134: nit: /NonExisting/NonExistent/

dasch-1:
- Created new class ExplicitRowTypeTable that extends from Table
- Add validation logic to SqlCreateTable that creates an instance of ExplicitRowTypeTable and adds it to the schema
- Added tests

PR #321, opened 2020-08-18
[CALCITE] Added functions created using CREATE FUNCTION to the schema

AndrewPochapsky:
- created class FunctionParameterImpl and UserDefinedFunction
- added validation method to SqlCreateFunctionSqlForm which adds function to schema
- Changed opTab of SqlValidatorImpl to be a ChainedSqlOperatorTable containing the passed in opTab and the catalogReader
- Updated ChainedSqlOperatorTable `of()` method to add all tables within any table of instance ChainedSqlOperatorTable


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3227,6 +3240,28 @@ public void validateIntervalQualifier(SqlIntervalQualifier qualifier) {
     }
   }
 
+  @Override public void validateCreateFunctionSqlForm(
+      SqlCreateFunctionSqlForm createFunction) {
+    Preconditions.checkArgument(createFunction != null);
+    nodeToTypeMap.put(createFunction, unknownType);
```
dstekol, line 3246: I think the usual behavior for validation is to call setValidatedNodeType rather than writing to the nodeToTypeMap directly.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3227,6 +3240,28 @@ public void validateIntervalQualifier(SqlIntervalQualifier qualifier) {
     }
   }
 
+  @Override public void validateCreateFunctionSqlForm(
+      SqlCreateFunctionSqlForm createFunction) {
+    Preconditions.checkArgument(createFunction != null);
```
dstekol, line 3245: What would happen if this method were called twice with the same function signature (if the user accidentally defined the same function twice, or wanted to define it and then overwrite its behavior)?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I don't think that's possible for this case since setValidatedNodeType doesn't allow the setting of unknownType

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I just tried it and it results in the "duplicate" call being added to the schema as well. Behavior wise the function which was added first gets chosen when matching. I'm not sure how much of an issue this is as usually a "DROP" call comes before redefining a structure right?

AndrewPochapsky: Though it should be possible to add the ability to "overwrite" entries that have an already matching signature.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dstekol: Hmm ok, why is unknownType being inserted instead of returnType?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dstekol: Not sure if this is an assumption we can rely on - @dasch-1, what is the expected behavior if the same function gets created twice?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I guess it's a question of semantics, I feel like a "CREATE" statement doesn't really have an inherent type. I don't mind changing it to returnType though.

dstekol: Oh, that's fair, CREATE doesn't necessarily have the return type of the function it is defining

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I think it makes sense to overwrite a function of the same signature if it is added again.

dstekol: I seem to remember you may get an error if you try to CREATE the same thing more than once in SQL.

dasch-1: Just to clarify, we're only talking about a CREATE FUNCTION not a REPLACE FUNCTION, right?

core/src/main/java/org/apache/calcite/schema/impl/FunctionParameterImpl.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.apache.calcite.schema.impl;
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.schema.FunctionParameter;
+
+import java.util.Objects;
+
+public class FunctionParameterImpl implements FunctionParameter {
```
dasch-1, line 25: nit: Javadoc missing for this class and its constructor.

core/src/main/java/org/apache/calcite/schema/impl/UserDefinedFunction.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.schema.FunctionParameter;
+import org.apache.calcite.schema.ScalarFunction;
+
+import java.util.List;
+import java.util.Objects;
+
+public class UserDefinedFunction implements ScalarFunction {
```
dasch-1, line 27: nit: Javadoc missing for this class and its constructor.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -306,8 +311,15 @@ protected SqlValidatorImpl(
       SqlValidatorCatalogReader catalogReader,
       RelDataTypeFactory typeFactory,
       Config config) {
-    this.opTab = Objects.requireNonNull(opTab);
     this.catalogReader = Objects.requireNonNull(catalogReader);
+    SqlOperatorTable readerTable = (SqlOperatorTable) catalogReader;
```
dasch-1, line 315: Is this explicit cast needed?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -306,8 +311,15 @@ protected SqlValidatorImpl(
...
+    if (opTab instanceof ChainedSqlOperatorTable) {
```
dasch-1, line 316: Rather than having this instanceof check and all of the casting logic here, we could handle this in ChainedSqlOperatorTable's `of()` method. Any operator tables that are specified in `of()` which are themselves of type ChainedSqlOperatorTable would have their individual tables extracted and added.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dstekol: I believe so.

core/src/main/java/org/apache/calcite/sql/util/ChainedSqlOperatorTable.java
```diff
@@ -43,12 +42,12 @@
    * Creates a table based on a given list.
    */
   public ChainedSqlOperatorTable(List<SqlOperatorTable> tableList) {
-    this.tableList = ImmutableList.copyOf(tableList);
+    this.tableList = new ArrayList<>(tableList);
```
dasch-1, line 45: Rather than relaxing the mutability of `tableList`, one option would be for callers to just construct a brand new ChainedSqlOperatorTable using the `of()` method if they wanted to add a new operator table.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dasch-1: Since two subsequent CREATE (but not necessarily CREATE OR REPLACE, or REPLACE) calls should be illegal, I think we can handle this in any reasonable way we want. Overriding the previous definition makes sense. Or throwing an error makes sense too.

But we probably don't want both function signatures to exist simultaneously in the catalog with only one that's accessible.

core/src/main/java/org/apache/calcite/sql/util/ChainedSqlOperatorTable.java
AndrewPochapsky: That was the initial approach but many tests failed as a result, I can look into it again though.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: As far as I can tell this is not the current behavior of the `of()` method. Are you suggesting that I should add this behaviour?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: We need it because catalogReader is of type SqlValidatorCatalogReader which isn't of type SqlOperatorTable. It is assumed however that the instance of catalog reader is of type CalciteCatalogReader, which implements Prepare.CatalogReader which implements SqlOperatorTable.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
AndrewPochapsky: I decided to go with the overwriting approach and added a test to demonstrate this.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dasch-1: Yep, that's what I'm suggesting. It seems like a caller could want to chain ChainedSqlOperatorTable's using the of() method for other use cases too.

core/src/main/java/org/apache/calcite/schema/impl/FunctionParameterImpl.java
```diff
@@ -0,0 +1,86 @@
...
+  private final int ordinal;
+  private final String name;
+  public final RelDataType type;
+  private final boolean optional;
+
+  /**
+   * Creates a {@code FunctionParameterImpl}.
+   *
+   * @param ordinal The index of the parameter
+   * @param name The name of the parameter
```
dasch-1, line 39: nit: "Must not be null" (for `type` parameter as well)

core/src/main/java/org/apache/calcite/schema/impl/FunctionParameterImpl.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Override public RelDataType getType(RelDataTypeFactory typeFactory) {
+    return type;
+  }
+
+  @Override public boolean isOptional() {
+    return optional;
+  }
+
+  @Override public int hashCode() {
```
dasch-1, line 67: Can you explain a bit what we're doing with the `hashCode()` and `equals()`? Has this been added to the design doc?

core/src/main/java/org/apache/calcite/schema/impl/UserDefinedFunction.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class UserDefinedFunction implements ScalarFunction {
+
+  private final List<FunctionParameter> parameters;
+  private final RelDataType returnType;
+
+  /**
+   * Creates a {@code UserDefinedFunction}.
+   *
+   * @param parameters The parameters of the created function
```
dasch-1, line 39: nit: "Must not be null" (same for `returnType` as well)

core/src/main/java/org/apache/calcite/util/NameMultimap.java
```diff
@@ -60,10 +82,21 @@ public NameMultimap() {
...
 
-  /** Adds an entry to this multimap. */
+  /**
+   * Adds an entry to this multimap. If allowsDuplicates is true, then the
+   * existing entry will be overwritten.
+   */
   public void put(String name, V v) {
     List<V> list = map().computeIfAbsent(name, k -> new ArrayList<>());
-    list.add(v);
+    int index = -1;
```
dasch-1, line 91: Why do we need to perform all of this? Can't we just use a Set rather than an Array data structure if duplicates aren't allowed?

core/src/main/java/org/apache/calcite/schema/impl/FunctionParameterImpl.java
AndrewPochapsky: I needed to add them so duplicates could be checked for.

core/src/main/java/org/apache/calcite/util/NameMultimap.java
AndrewPochapsky: I'll try changing it to a Set.

core/src/main/java/org/apache/calcite/util/NameMultimap.java
```diff
@@ -87,35 +88,34 @@ public NameMultimap(boolean allowsDuplicates) {
...
-    if (index >= 0) {
-      list.set(index, v);
+    Collection<V> collection = null;
+    if (allowsDuplicates) {
+      collection = map().computeIfAbsent(name, k -> new ArrayList<>());
     } else {
-      list.add(v);
+      collection = map().computeIfAbsent(name, k -> new HashSet<>());
+      // Remove old instance so it gets overridden.
+      collection.remove(v);
```
dasch-1, line 97: Why do we need to explicitly remove here? Shouldn't the HashSet automatically handle the overwriting?

AndrewPochapsky: I'm pretty sure if an item is already in the set it just doesn't get added, therefore removing it first ensures that it gets overwritten.

dasch-1: Backing up a little bit: why are we editing the NameMultimap class at all? As the name suggests, duplicates should be always be allowed.

AndrewPochapsky: From my knowledge NameMultimap is used for functions because it can store overloads of a function (which are not duplicates in the sense that I am defining them to be). If we used NameMap, we would not be able to support overloads.

dasch-1: Ah I see, so Multimap just refers to allowing multiple functions with the same name but different parameter lengths?

This seems like a fair amount of changes to add support for duplicate functions (as in same name and number of parameters). Would it be simpler just to throw an error if a duplicate function is detected rather than allowing for it to happen?

AndrewPochapsky: From a glance the amount of changes seem similar since allowing duplicates is the default behavior so the custom boolean would still be required. However, we could avoid needing to make the List a collection probably. My main concern is that I don't think that the schema resets during tests so some additional work would need to go into exposing a reset mechanism (this may be trivial but I haven't take a look yet).

dasch-1: Based on the test in UtilTest::testNameMultimap, it appears that the intention of this class is to explicitly allow duplicates of whatever object it is storing. By making these proposed changes, it feels like we're kind of abusing this original intent.

What if we instead stored CalciteSchema's FunctionEntry objects in a NameMap? And we updated the hashCode() and equals() methods for the FunctionEntryImpl (or something similar) to be able to differentiate two functions with the same name but different argument lengths? Then we would automatically get deduplication for free from that data structure.

AndrewPochapsky: Is this what you had in mind? https://github.com/AndrewPochapsky/calcite/compare/function-schema...AndrewPochapsky:name-map?expand=1

To me this approach looks equivalent to using a NameMultimap minus the abstraction.

PR #323, opened 2020-08-18
[CALCITE] Unknown compound identifiers

dstekol:
When tables are specified via compound identifiers like foo.bar, this will now result in prefixes getting added as subschemas to the root schema, and tables getting added to the appropriate subschemas (ex. "foo" will be added as a subschema to the root, and "bar" will be added as a table to subschema "foo").

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
@@ -22,7 +22,6 @@ import org.apache.calcite.prepare.RelOptTableImpl;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.schema.Table;
-import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.calcite.schema.impl.UnknownPlaceholderTable;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlIdentifier;
@@ -206,7 +205,7 @@ public class IdentifierNamespace extends AbstractNamespace {
    * @return    A placeholder {@code TableNamespace}.
    */
   private SqlValidatorNamespace getPlaceholderNamespace(SqlIdentifier id) {
-    CalciteSchema parentSchema = getOrCreateParentSchema(id);
+    CalciteSchema parentSchema = validator.getOrCreateParentSchema(id);
     SqlIdentifier tableId = id.getComponent(id.names.size() - 1);
     Table table = new UnknownPlaceholderTable(validator.getTypeFactory());
     SqlValidatorTable valTable = RelOptTableImpl.create(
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/apache/calcite/test/SqlValidatorBestEffortTest.java
```diff
@@ -91,20 +91,20 @@
   @Test public void testUnknownTableMerge() {
     String sql = "MERGE INTO foo AS f USING dept AS b ON f.bar = b.name"
```
dstekol, line 92: Since UnknownRecordType objects are now connected to an underlying schema, each time a field is queried within them, that field gets stored within the record (in other words, if you have an unknown table and you query whether it contains column abc, the table will return yes, and will internally record that one of the fields it contains is column abc). This meant that I had to tweak existing tests to account for this behavior.

core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -204,11 +206,42 @@ private SqlValidatorNamespace resolveImpl(SqlIdentifier id) {
...
+   * schema, bar will be added as a subschema to foo, and then bar will be
+   * returned as the parent schema of baz.
+   * @param id  the identifier for which the parent schema must be created or
+   *            retrieved.
+   * @return    the immediate parent of the object denoted by the identifer -
+   *            in other words, the schema corresponding to the second to last
+   *            component of the identifier. If the identifier consists of only
+   *            a single name, the root schema will be returned.
+   */
+  public CalciteSchema getOrCreateParentSchema(SqlIdentifier id) {
```
AndrewPochapsky, line 233: Is it possible to move this to SqlValidator? I'm not sure if I'll be able to access this from there otherwise.

dstekol: Ok, just moved it.

AndrewPochapsky: Awesome, thanks!

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3976,6 +3978,34 @@ public SqlValidatorScope getWithScope(SqlNode withItem) {
...
+   *            retrieved.
+   * @return    the immediate parent of the object denoted by the identifer -
+   *            in other words, the schema corresponding to the second to last
+   *            component of the identifier. If the identifier consists of only
+   *            a single name, the root schema will be returned.
+   */
+  @Override public CalciteSchema getOrCreateParentSchema(SqlIdentifier id) {
+    CalciteSchema parentSchema =
+        getCatalogReader().getRootSchema();
+    //add chain of subschemas corresponding to identifier prefixes
```
dasch-1, line 3998: nit: `// Add chain ...`

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3976,6 +3978,34 @@ public SqlValidatorScope getWithScope(SqlNode withItem) {
...
+    for (int i = 0; i < id.names.size() - 1; i++) {
+      //check that subschema is not already present before adding
```
dasch-1, line 4000: nit: `// Check that ...`

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3976,6 +3978,34 @@ public SqlValidatorScope getWithScope(SqlNode withItem) {
...
+      if (parentSchema.getSubSchema(id.names.get(i), false) == null) {
```
dasch-1, line 4001: nit: Use inline comments to indicate what `false` represents.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3976,6 +3978,34 @@ public SqlValidatorScope getWithScope(SqlNode withItem) {
...
+        parentSchema.add(id.names.get(i), new AbstractSchema());
+      }
+      parentSchema = parentSchema.getSubSchema(id.names.get(i), false);
```
dasch-1, line 4004: nit: Use inline comments to indicate what `false` represents.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dasch-1, line 4001: nit: Since it's referenced a few times in this function, let's store `id.names.get(i)` in a local variable.

PR #324, opened 2020-08-18
[CALCITE] Add core validation in CREATE PROCEDURE statements

michaelvdow:
Adds support for validating SELECT, INSERT, MERGE, UPDATE, and DELETE statements that reference table columns inside CREATE PROCEDURE statements.


core/src/main/java/org/apache/calcite/sql/SqlCreateProcedure.java
```diff
@@ -105,7 +105,12 @@ public SqlCreateProcedure(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
 
   @Override public void validate(SqlValidator validator,
       SqlValidatorScope scope) {
-    if (statement instanceof SqlScriptingNode) {
+    if (statement instanceof SqlScriptingNode
```
dasch-1, line 108: Since this logic looks identical to validateSqlNodeList in SqlScriptingNode, should we move it to a common location (e.g. SqlValidatorImpl) and call it from both locations?

michaelvdow: Yeah that sounds good. I'll add another method to move this logic out

core/src/main/java/org/apache/calcite/sql/SqlCreateProcedure.java
```diff
@@ -105,12 +105,7 @@ public SqlCreateProcedure(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
 
   @Override public void validate(SqlValidator validator,
       SqlValidatorScope scope) {
-    if (statement instanceof SqlScriptingNode
-        || statement instanceof SqlSelect
-        || statement instanceof SqlDelete
-        || statement instanceof SqlInsert
-        || statement instanceof SqlMerge
-        || statement instanceof SqlUpdate) {
+    if (validator.shouldValidateInCreateProcedure(statement)) {
```
dasch-1, line 108: nit: We might even be able to move even more logic into the shared SqlValidator function.

What if SqlCreateProcedure's validate() method just became:
`validator.validateScriptingStatement(statement, scope);`

And SqlScriptingNode's validate() method became:
`for (SqlNode node : nodes) {
  validator.validateScriptingStatement(node, scope);
}`

And then the if-check and the actual call to validate() could all be handled within SqlValidator's validateScriptingStatement() method.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1185,6 +1186,15 @@ public SqlValidatorScope getOverScope(SqlNode node) {
     return scopes.get(node);
   }
 
+  @Override public boolean shouldValidateInCreateProcedure(SqlNode node) {
+    return (node instanceof SqlScriptingNode
```
dasch-1, line 1190: These extra parentheses around the return statement appear to be causing the linter check to fail.

core/src/main/java/org/apache/calcite/sql/SqlCreateProcedure.java
michaelvdow: Ah that's true, I like this method better. I'll update and fix the parentheses issue too

PR #330, opened 2020-08-19
[CALCITE] Fix column expansion from unknown subqueries

dstekol:
Columns deriving from unknown subqueries no longer be rewritten with the ** prefix.


core/src/main/java/org/apache/calcite/rel/type/RelDataTypeImpl.java
```diff
@@ -112,7 +112,12 @@ public RelDataTypeField getField(String fieldName, boolean caseSensitive,
     for (RelDataTypeField field : fieldList) {
       if (field.isDynamicStar()) {
         // the requested field could be in the unresolved star
-        return field;
+        return new RelDataTypeFieldImpl(
+            fieldName,
+            field.getIndex(),
+            new BasicSqlType(((BasicSqlType) field.getValue()).typeSystem, SqlTypeName.ANY,
+                /*nullable=*/true, /*precision=*/-1, /*scale=*/-1,
```
michaelvdow, line 119: nit: Add space after inline comment

core/src/test/java/org/apache/calcite/test/SqlValidatorBestEffortTest.java
```diff
@@ -230,6 +230,29 @@
         .rewritesTo(expected);
   }
 
+  @Test public void testSelectWithUnknownSubquery() {
+    String sql = "with t AS (select * FROM foo) select (select 2 AS x from t where x = 1) from "
```
michaelvdow, line 234: nit: Line wrap

dasch-1:
FYI, you'll also likely need to revert the following change made to an existing unit test to allow it to work with the "CATALOG" bug:

https://github.com/googleinterns/calcite/pull/332/files#diff-e4c6c1f081085f49ec72793dadf18276R316

PR #335, opened 2020-08-20
[CALCITE] CREATE <macro> statements get added to schema 

AndrewPochapsky:
- Added macroMap field to CalciteSchema
- Added Macro class
- Refactored logic that computes the schema list within `getFunctionsFrom(...)` into a separate function as it is also used for this feature
- Added Macros to the existing function validation infrastructure


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1338,16 +1340,21 @@ protected SqlNode performUnconditionalRewrites(
...
-        final List<SqlOperator> overloads = new ArrayList<>();
-        opTab.lookupOperatorOverloads(function.getNameAsId(),
-            function.getFunctionType(), SqlSyntax.FUNCTION, overloads,
-            catalogReader.nameMatcher());
-        if (overloads.size() == 1) {
-          ((SqlBasicCall) call).setOperator(overloads.get(0));
+        SqlFunctionCategory category = function.getFunctionType();
+        // Don't want to overwrite macros since they lose their identifiable
+        // category.
+        if (category == null || !category.isUserDefinedMacro()) {
```
AndrewPochapsky, line 1346: This is necessary because `lookupOperatorOverloads` will convert this a SqlUserDefinedFunction with category USER_DEFINED_FUNCTION. Ideally the category of this one is set to USER_DEFINED_MACRO on creation, but the constructor that allows that is protected: https://github.com/googleinterns/calcite/blob/97399801253a1417ff2676c8740c06180b2e26f1/core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java#L56.

Is it alright to make this public?

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -137,8 +139,53 @@ public CalciteCatalogReader withSchemaPath(List<String> schemaPath) {
...
+   *
+   * @return The found macro, or null
+   */
+  public Macro getMacro(List<String> names) {
+    final List<List<String>> schemaNameList = computeSchemaNameList(names);
+    for (List<String> schemaNames : schemaNameList) {
+      CalciteSchema schema =
+          SqlValidatorUtil.getSchema(rootSchema,
+              Iterables.concat(schemaNames, Util.skipLast(names)), nameMatcher);
+      if (schema != null) {
```
dasch-1, line 156: nit: Unnest by reversing polarity and continuing instead.

dasch-1:
Making that SqlUserDefinedFunction constructor public sounds ok to me.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -137,8 +139,53 @@ public CalciteCatalogReader withSchemaPath(List<String> schemaPath) {
...
+        final String name = Util.last(names);
+        boolean caseSensitive = nameMatcher.isCaseSensitive();
+        functions2.addAll(schema.getFunctions(name, caseSensitive));
+        Macro macro = schema.getMacro(name, caseSensitive);
```
dasch-1, line 179: Why does this call `getMacro()` in a function about getting functions?

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -137,8 +139,53 @@ public CalciteCatalogReader withSchemaPath(List<String> schemaPath) {
...
 
+  /**
+   * Gets the macro with the given name if it exists.
+   *
+   * @param names The name of the macro, list represents possibility of a
+   *              compound identifier
+   *
+   * @return The found macro, or null
+   */
+  public Macro getMacro(List<String> names) {
```
dasch-1, line 150: This looks like it has a lot of duplicate code to `getFunctionsFrom()`. Is there a way refactor shared logic out into a common helper function that they can both call?

core/src/main/java/org/apache/calcite/sql/util/ReflectiveSqlOperatorTable.java
```diff
@@ -89,7 +89,9 @@ public void lookupOperatorOverloads(SqlIdentifier opName,
       SqlFunctionCategory category, SqlSyntax syntax,
       List<SqlOperator> operatorList, SqlNameMatcher nameMatcher) {
     // NOTE jvs 3-Mar-2005:  ignore category until someone cares
-
+    if (category != null && category.isUserDefinedMacro()) {
```
dasch-1, line 92: Why do we need this new check? Can you add a comment explaining?

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -260,6 +298,10 @@ public void lookupOperatorOverloads(final SqlIdentifier opName,
       predicate = function ->
           function instanceof TableMacro
               || function instanceof TableFunction;
+    } else if (category.isUserDefinedFunction()) {
```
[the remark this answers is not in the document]
AndrewPochapsky: If we don't have it then potential functions for EXEC calls will also include UDFs which is invalid.

core/src/main/java/org/apache/calcite/schema/Macro.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.calcite.schema;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Schema class for all user defined macros.
+ */
+public class Macro {
```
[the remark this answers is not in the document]
AndrewPochapsky: I don't think so as that is used for macros that return tables. It also an interface that has a [function apply](https://github.com/googleinterns/calcite/blob/97399801253a1417ff2676c8740c06180b2e26f1/core/src/main/java/org/apache/calcite/schema/TableMacro.java#L35) that doesn't make sense for this case.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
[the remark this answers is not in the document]
AndrewPochapsky: Actually I meant the reverse, not having this would allow macros to be called without EXEC. However I just thought of a better approach.

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
```diff
@@ -81,6 +84,7 @@ protected CalciteSchema(CalciteSchema parent, Schema schema,
     this.parent = parent;
     this.schema = schema;
     this.name = name;
+    this.macroMap = new NameMap<>();
```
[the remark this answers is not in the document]
AndrewPochapsky: I can add that but since we aren't currently passing in a macroMap in the constructor I didn't think it was necessary.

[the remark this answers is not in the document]
dasch-1: Leaving it is sounds good then.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
dasch-1: SGTM

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
```diff
@@ -220,6 +223,16 @@ public FunctionEntry add(String name, Function function) {
     return entry;
   }
 
+  public FunctionEntry add(String name, Macro macro) {
```
dasch-1, line 226: Can multiple macros with the same name but different number of parameters be defined?

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -137,8 +138,52 @@ public CalciteCatalogReader withSchemaPath(List<String> schemaPath) {
...
 
+  /**
+   * Gets the macro with the given name if it exists.
+   *
+   * @param names The name of the macro, list represents possibility of a
+   *              compound identifier
+   *
+   * @return The found macro, or null
+   */
+  private Macro getMacro(List<String> names) {
```
dasch-1, line 149: This still looks very similar to `getFunctionsFrom()`. Is there a way to refactor common logic into a helper function?

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
AndrewPochapsky: They can't as far as I know.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
AndrewPochapsky: I'm not sure what else could be refactored, do you have any ideas?

core/src/main/java/org/apache/calcite/schema/Macro.java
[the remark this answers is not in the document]
dasch-1: I see. So is a TableMacro a subset of a Macro? It seems like these two classes should be linked by inheritance in some way to show their relationship.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
dasch-1: If macros can't have duplicate names like functions, then probably nothing of value.

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
dasch-1: Ok sounds good.

core/src/main/java/org/apache/calcite/schema/Macro.java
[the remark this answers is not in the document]
AndrewPochapsky: Since TableMacro is an interface I'm not sure how to connect it up to Macro (which is not an interface). We could make Macro an interface itself but that seems unnecessary since Macro doesn't itself have any interesting properties, it's just there so we can differentiate it between other Function implementations. What do you think?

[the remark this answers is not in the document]
dasch-1: Makes sense. We can leave them as is then.

core/src/main/java/org/apache/calcite/schema/impl/Macro.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.calcite.schema.Function;
+import org.apache.calcite.schema.FunctionParameter;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Schema class for all user defined macros.
+ */
+public class Macro implements Function {
```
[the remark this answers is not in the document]
dasch-1: Answered in another comment thread.

PR #346, opened 2020-08-25
[CALCITE] Extraction of procedure schema

AndrewPochapsky:
Added extraction of create procedure signature and hooked up call statements to the function validation infrastructure.


core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -138,6 +139,31 @@ public CalciteCatalogReader withSchemaPath(List<String> schemaPath) {
...
 
+  /**
+   * Gets the procedure with the given name if it exists.
+   *
+   * @param names The name of the procedure, list represents possibility of a
+   *              compound identifier
+   *
+   * @return The found procedure, or null
+   */
+  private Procedure getProcedure(List<String> names) {
```
dasch-1, line 150: This looks identical to `getMacro()`. Can we condense these two cases into a single function?

core/src/main/java/org/apache/calcite/schema/impl/Procedure.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Schema class for all user defined procedures.
+ */
+public class Procedure implements Function {
+  private final List<FunctionParameter> parameters;
+
+  /**
+   * Creates a {@code Procedure}.
+   *
+   * @param parameters The procedure parameters
```
dasch-1, line 34: nit: Must not be null

core/src/main/java/org/apache/calcite/sql/SqlFunctionCategory.java
```diff
@@ -65,8 +67,8 @@ public boolean isUserDefined() {
     return properties.contains(USER_DEFINED);
   }
 
-  public boolean isUserDefinedFunction() {
```
dasch-1, line 68: Why are we deleting this existing function? Is it not used anymore?

core/src/main/java/org/apache/calcite/sql/SqlCreateProcedure.java
```diff
@@ -105,6 +105,7 @@ public SqlCreateProcedure(SqlParserPos pos, SqlCreateSpecifier createSpecifier,
 
   @Override public void validate(SqlValidator validator,
       SqlValidatorScope scope) {
+    validator.addProcedureToSchema(this);
```
dasch-1, line 108: Why not follow the pattern used for other nodes that call back to SqlValidatorImpl with something like `validate<NodeType>()`? In this case, that would be `validateCreateProcedure()`. That function could take care of adding the procedure to the schema and validating the underlying scripting statement.

parsing/dialects/intermediate/dialects/dialect1/src/test/java/org/apache/calcite/test/Dialect1ValidatorTest.java
```diff
@@ -240,6 +240,50 @@ public SqlNode parseAndValidate(String sql) {
...
+        + "return 1";
+    String query = "CALL foo(1)";
+    String query2 = "SELECT foo(1)";
+    sql(procedureDdl).ok();
+    sql(functionDdl).ok();
+    sql(query).type("ANY NOT NULL");
+    sql(query2).type("RecordType(INTEGER NOT NULL EXPR$0) NOT NULL");
+  }
+
+  @Test public void testCreateProcedureSchemaInvalidNumberFails() {
```
dasch-1, line 272: `testCreateProcedureSchemaInvalidNumberOfParamsFails`

core/src/main/java/org/apache/calcite/sql/SqlFunctionCategory.java
AndrewPochapsky: Yeah, I was initially using it in the previous feature but forgot to remove it.

core/src/main/java/org/apache/calcite/sql/SqlCreateProcedure.java
AndrewPochapsky: Sure, that makes more sense.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -156,9 +171,17 @@ private Procedure getProcedure(List<String> names) {
...
+      Function function = null;
+      switch (category) {
+      case USER_DEFINED_PROCEDURE:
+        function = schema.getProcedure(name, caseSensitive);
+        break;
+      case USER_DEFINED_MACRO:
+        function = schema.getMacro(name, caseSensitive);
+        break;
+      }
+      if (function != null) {
```
dasch-1, line 183: nit: Can we remove this and just `return function` unconditionally at the end of this function?

AndrewPochapsky: I'm not 100% sure on this one. All tests pass if I return function at the end but I don't know if that will work for all cases. Could it happen that function is not null and then later becomes null via more iterations of the loop?

dasch-1: Ah I see what you mean. In that case, let's just leave it as is.

PR #350, opened 2020-08-27
[CALCITE] Allow BYTE type in TRIM function

AndrewPochapsky:
- Added parsing logic of byte literals to dialect1
- Made small adjustments to validation logic of TRIM function to allow for byte types
- Added big query specific byte literal unparsing logic


core/src/main/java/org/apache/calcite/sql/SqlHexCharStringLiteral.java
```diff
@@ -33,16 +34,30 @@
   final HexCharLiteralFormat format;
   final CharacterSet charSet;
 
+  public SqlHexCharStringLiteral(final String hex, final SqlParserPos pos,
```
AndrewPochapsky, line 37: Now that I look at this more I think it makes more sense to create a new class to deal with bytes, thoughts?

AndrewPochapsky: Especially with the differences between unparsing a byte literal and a regular hex string.

dasch-1: Yes, I think a separate byte literal class sounds like a good idea.

AndrewPochapsky: I've made the changes.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -290,7 +306,7 @@ private void unparseTrim(SqlWriter writer, SqlCall call, int leftPrec,
     // If the trimmed character is a non-space character, add it to the target SQL.
     // eg: TRIM(BOTH 'A' from 'ABCD'
     // Output Query: TRIM('ABC', 'A')
-    if (!valueToTrim.toValue().matches("\\s+")) {
+    if (!valueToTrim.toValue().matches("\\s*")) {
```
dasch-1, line 309: Why is this changed needed?

core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
```diff
@@ -154,6 +155,25 @@ public SqlCall createCall(
...
+   * Returns either a padding char literal or an empty byte literal depending
+   * on the type of {@code node} passed in.
+   *
+   * @param node The node whose type is checked
+   * @param pos The parser position
+   *
+   * @return The null literal
+   */
+  private static SqlLiteral getNullLiteral(SqlNode node, SqlParserPos pos) {
+    if (node instanceof SqlLiteral) {
```
dasch-1, line 168: nit: Reduce nesting by reversing polarity and returning `createCharString` early.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -103,6 +97,64 @@ public static boolean isCharTypeComparable(List<RelDataType> argTypes) {
...
+   * Checks whether two types or more are byte comparable.
+   *
+   * @return Returns true if all operands are of byte type.
+   */
+  public static boolean isByteTypeComparable(List<RelDataType> argTypes) {
+    assert argTypes != null;
+    assert argTypes.size() >= 2;
+    List<RelDataType> argTypes2 = filterAnyAndNull(argTypes);
+    for (Pair<RelDataType, RelDataType> pair : Pair.adjacents(argTypes2)) {
+      RelDataType t0 = pair.left;
```
dasch-1, line 110: nit: Can probably remove these intermediate variables and just reference `pair.left` and `pair.right` directly below.

parsing/dialects/intermediate/dialects/dialect1/parserImpls.ftl
```diff
@@ -3754,6 +3769,34 @@ SqlDrop SqlDrop() :
...
+ * is an SqlCall expression, which concatenates 2 or more string
+ * literals; the validator reduces this.
+ */
+SqlNode Literal() :
+{
+    SqlNode e;
+}
+{
+    (
+        e = NumericLiteral()
```
dasch-1, line 3784: nit: While we're here, can we alphabetically order these clauses?

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
AndrewPochapsky: Since the null byte string is empty not having this would result in TRIM(<byte>, b''). Is this valid?

dasch-1: Is this different than if the argument was an empty hex char string in TRIM()?

AndrewPochapsky: It is in the case of byte since by the documentation, if a byte is passed in the default is a null byte.

dasch-1: Can you paste a snippet of the text from the docs that you're referring to?

AndrewPochapsky: If you omit this option (refers to BOTH/TRAILING/LEADING), the default is BOTH, and the default trim character is a null byte for byte types and a pad character for character types.

dasch-1: Ok I think I see what you mean. So this is checking whether the character to trim is a space character or not? Not the string to be trimmed.

AndrewPochapsky: Yeah exactly.

dasch-1: Ok in that case, let's just update the comment above to indicate that it should be a `non-space character and non-empty`.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -147,6 +148,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
     unparseFetchUsingLimit(writer, offset, fetch);
   }
 
+  @Override public void unparseByteLiteral(SqlWriter writer,
+      SqlByteLiteral literal, int leftPrec, int rightPrec) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("b'");
+    String hexString = literal.getValue().toString();
+    // Start at 1 and end at length - 1 to remove surrounding single quotes.
```
lisayin, line 156: Do we need to worry about whitespace?

AndrewPochapsky: Good question. I don't think we need to worry about it since the parsing logic won't allow something like that to go through since a white space isn't a hex digit. I can add another comment saying that we are assuming that there are no whitespaces though.

